PR #48, opened 2024-10-17
Update book-library to the latest version from JS3

illicitonion:
I'm not sure how it ended up as weirdly broken as it did, but I'm pretty sure it wasn't intended.


SallyMcGrath (approved):
WHOOPS we missed this

PR #53, opened 2024-12-02
London |Elhadj Abdoul Diallo | Module-Data-Flows| WEEK1

eediallo:
## Learners, PR Template

Self checklist

## Changelist

I have done the three exercises in destructuring

## Questions

### DIFF
--- a/Sprint-1/destructuring/exercise-1/exercise.js
+++ b/Sprint-1/destructuring/exercise-1/exercise.js
@@ -4,8 +4,7 @@ const personOne = {
   favouriteFood: "Spinach",
 };
 
-// Update the parameter to this function to make it work.
-// Don't change anything else.
+
 function introduceYourself({name, age, favouriteFood} = personOne) {
   console.log(
     `Hello, my name is ${name}. I am ${age} years old and my favourite food is ${favouriteFood}.`
--- a/Sprint-1/destructuring/exercise-1/exercise.js
+++ b/Sprint-1/destructuring/exercise-1/exercise.js
@@ -6,7 +6,7 @@ const personOne = {
 
 // Update the parameter to this function to make it work.
 // Don't change anything else.
-function introduceYourself(___________________________) {
+function introduceYourself({name, age, favouriteFood} = personOne) {
   console.log(
     `Hello, my name is ${name}. I am ${age} years old and my favourite food is ${favouriteFood}.`
   );
--- a/Sprint-1/destructuring/exercise-3/exercise.js
+++ b/Sprint-1/destructuring/exercise-3/exercise.js
@@ -8,11 +8,11 @@ let order = [
 ];
 
 
-function orderReceipt(){
+function orderReceipt(arr){
   
   console.log('QTY'.padEnd(8),'ITEM'.padEnd(18),'TOTAL')
   let totalOrder = 0
-  for(const {quantity, itemName, unitPricePence} of order){
+  for(const {quantity, itemName, unitPricePence} of arr){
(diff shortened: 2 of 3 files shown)


Sprint-1/destructuring/exercise-3/exercise.js
```diff
@@ -6,3 +6,19 @@ let order = [
...
+  
+  console.log('QTY'.padEnd(8),'ITEM'.padEnd(18),'TOTAL')
+  let totalOrder = 0
+  for(const {quantity, itemName, unitPricePence} of arr){
+    const totalPerItem = ((unitPricePence * quantity) / 100).toFixed(2)
+    totalOrder += unitPricePence * quantity
+    console.log(`${quantity.toString().padEnd(8)}${itemName.padEnd(20)}${totalPerItem}`)
+  }
+
+  console.log(`\nTotal: ${(totalOrder/100).toFixed(2)}`)
```
illicitonion, line 21: There are a few pieces of repetitive code here (code where you write down how to do the exact same thing twice, where having a name for that operation could also be clarifying) - can you identify and deduplicate them?

Sprint-1/destructuring/exercise-1/exercise.js
```diff
@@ -4,9 +4,8 @@ const personOne = {
   favouriteFood: "Spinach",
 };
 
-// Update the parameter to this function to make it work.
-// Don't change anything else.
-function introduceYourself(___________________________) {
+
+function introduceYourself({name, age, favouriteFood} = personOne) {
```
illicitonion, line 8: What does the `= personOne` do here? Why is it needed?

illicitonion (commented):
Looking good, congrats!

Sprint-1/destructuring/exercise-1/exercise.js
eediallo: ```=personOne``` acts as the default array here, meaning if no other array is passed the values ```{name, age, favouriteFood}``` will come from it. It is needed here because we are only allowed to change the parameter of the function nothing else.

illicitonion: What would have happened if you hadn't put the `= personOne` there?

eediallo: It would have worked fine!

illicitonion: Let's remove it then! :)

Sprint-1/destructuring/exercise-3/exercise.js
```diff
@@ -6,3 +6,24 @@ let order = [
   { itemName: "Hot Coffee", quantity: 2, unitPricePence: 100 },
   { itemName: "Hash Brown", quantity: 4, unitPricePence: 40 },
 ];
+
+const formatCurrency = (total) => (total / 100).toFixed(2);
+
+function formatReceiptDetails(str1, pad1Num, str2, pad2Num, str3) {
```
illicitonion, line 12: Every place you call this, you pass the same values for the pads - I would argue these probably shouldn't be parameters to the function, but should probably be constants that the function always uses?

(If you were using this function to format different styles of receipts, you may have this extra-parametrisable function, but you'd probably have a wrapper function setting the values consistently for each style of receipt you were formatting)

eediallo: Yes, I agree. The values are the same, therefore would make sense to set them as constants

illicitonion (commented):
Perfect!

The only feedback I have left is to be consistent with your formatting - some places you have a space between `for` and `(`, other places it's written `for(`. Prettier can help here :)

PR #59, opened 2024-12-11
LON|Amirhossein Aminian|Data Flows Module|Programmer Humour

Amir200524:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions


fetch/programmer-humour/javascript.js
```diff
@@ -0,0 +1,39 @@
...
+    .catch(error => {
+      // Handle any errors
+      console.error('There was an error!', error);
+      document.getElementById('loading').style.display = 'none'; // Hide loading message
+      document.getElementById('error-message').textContent = 'Failed to load comic. Please try again later.';
+    });
+}
+
+// Call the function to fetch the latest comic
+fetchLatestComic();
```
cjyuan, line 39: It is better to wrap the code to be executed on page load inside a "on window fully loaded" event listener.
```
window.addEventListener('load', function() {
  fetchLatestComic();
});
```

This way, the `<script>` tag can be safely placed anywhere in the HTML (with or without the `defer` property).

fetch/programmer-humour/javascript.js
```diff
@@ -0,0 +1,39 @@
...
+    })
+    .then(data => {
+      // Log the received data to the console
+      console.log(data);
+
+      // Get the image URL from the response data
+      const imageUrl = data.img;
+
+      // Render the image in the DOM
+      const comicImage = `<img src="${imageUrl}" alt="Latest XKCD Comic">`;
```
cjyuan, line 24: Since the value of `imageUrl` is coming from another website, there is a danger that its value can corrupt the generated string. (For details, you can ask ChatGPT `In JS, why is constructing HTML tag using template string is dangerous?`)

It is a safer practice to create `` element (or any other HTML element) using the `document.createElement()` approach.

cjyuan (commented):
Code is good. I only have a few suggestions.

PR #62, opened 2024-12-13
CYF London 11 | Nadika Zavodovska | Module-Data-Flows | Week 1 | Destructuring

nadika-zavodovska:
Self checklist

## Changelist

I have completed the Sprint 1 Destructuring exercise.

Thank you.

### DIFF
--- a/Sprint-1/destructuring/exercise-2/exercise.js
+++ b/Sprint-1/destructuring/exercise-2/exercise.js
@@ -70,3 +70,10 @@ let hogwarts = [
     occupation: "Teacher",
   },
 ];
+
+function createListGryffindor (hogwarts){
+  const listGryffindor = hogwarts.filter(({ house }) => house === "Gryffindor").map(({ firstName, lastName }) => `${firstName} ${lastName}`); 
+  
+}
+
+createListGryffindor(hogwarts);
--- a/Sprint-1/destructuring/exercise-2/exercise.js
+++ b/Sprint-1/destructuring/exercise-2/exercise.js
@@ -76,4 +76,10 @@ function createListGryffindor(hogwarts) {
   listGryffindor.forEach((name) => console.log(name));
 }
 
+function createListTeacherWithPet(hogwarts) {
+  const listTeacherWithPet = hogwarts.filter(({ pet, occupation }) => occupation === "Teacher" && pet !== null).map(({ firstName, lastName }) => `${firstName} ${lastName}`);
+  listTeacherWithPet.forEach((name) => console.log(name));
+}
+
 createListGryffindor(hogwarts);
+createListTeacherWithPet(hogwarts);
--- a/Sprint-1/destructuring/exercise-2/exercise.js
+++ b/Sprint-1/destructuring/exercise-2/exercise.js
@@ -71,9 +71,9 @@ let hogwarts = [
   },
 ];
 
-function createListGryffindor (hogwarts){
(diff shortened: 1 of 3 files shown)


Sprint-1/destructuring/exercise-2/exercise.js
```diff
@@ -70,3 +70,16 @@ let hogwarts = [
     occupation: "Teacher",
   },
 ];
+
+function createListGryffindor(hogwarts) {
+  const listGryffindor = hogwarts.filter(({ house }) => house === "Gryffindor").map(({ firstName, lastName }) => `${firstName} ${lastName}`);
```
tyzia, line 75: You are creating this function and parameters. What do you think createListGryffindor() will accept as parameters? Sure, when you will invoke this method, you will pass hogwarts to it (as this variable is defined for you), but it is up to you to come up with a good name for the parameters of your function. Think what this parameter will hold as a value and name it accordingly. This is not a show-stopper for your PR but it is good to practice.
Next question, what will happen if I will do something like this:

createListGryffindor(null);

Sprint-1/destructuring/exercise-3/exercise.js
```diff
@@ -6,3 +6,17 @@ let order = [
...
+
+function printOrder(order){
+  console.log(`${'QTY'.padEnd(8)}${'ITEM'.padEnd(20)}TOTAL`);
+
+  let total = 0;
+  order.forEach(({itemName, quantity, unitPricePence}) => {
+    const itemTotalPrice = unitPricePence / 100 * quantity;
+    total += itemTotalPrice;
+    console.log(`${quantity.toString().padEnd(8)}${itemName.padEnd(20)}${itemTotalPrice.toFixed(2)}`);
+  });
```
tyzia, line 18: You are applying correct approach when you need to iterate over the whole array and to count total value. But, there is a better option for the same task. Again, this doesn't mean that your code is not valid, but do you know what is the other way to achieve the same goal?

Sprint-1/destructuring/exercise-2/exercise.js
nadika-zavodovska: Thank you @tyzia for reviewing my task.
Yes, you are absolutely right. It's better to use more general names for a function's parameters. I changed the parameter to 'memberList.'

If I call the function with a `null` argument, I'll get a TypeError: Cannot read properties of null.
I added a condition to check if the parameter of the function is an array.

Sprint-1/destructuring/exercise-3/exercise.js
nadika-zavodovska: function printOrder(order) {
  console.log(`${'QTY'.padEnd(8)}${'ITEM'.padEnd(20)}TOTAL`);

  const total = order.reduce((runningTotal, { itemName, quantity, unitPricePence }) => {
    const itemTotalPrice = unitPricePence / 100 * quantity;
    runningTotal += itemTotalPrice;
    console.log(`${quantity.toString().padEnd(8)}${itemName.padEnd(20)}${itemTotalPrice.toFixed(2)}`);
    return runningTotal;
  }, 0);

  console.log(`\nTotal: ${total.toFixed(2)}`);
}

I can see, there is another solution with the reduce() method. So, the reduce() will do 2 steps in one: loop through the array and calculate the total. It's a clearer and easier way to handle this, thank you for asking me this question. It made me think in another way.

tyzia (approved):
Thank you for the update. Keep up good work!

nadika-zavodovska:
Thank you, Andrei, for your review, feedback, and support.

PR #94, opened 2024-12-23
Some fixes to the book project

illicitonion:
* Fix some not glorious naming ("let cell1 = row.insertCell(0);")
* Use innerText not innerHTML


SallyMcGrath (approved):
TY!

PR #99, opened 2024-12-26
CYF London | Fatma Arslantas | Module-Data-Flows | Week 1 | Array Destructuring

AFatmaa:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-1/destructuring/exercise-2/exercise.js
+++ b/Sprint-1/destructuring/exercise-2/exercise.js
@@ -73,9 +73,15 @@ let hogwarts = [
 
 // Task 1:
 // Function to display names of people in Gryffindor house
-function displayGryffindorMembers(hogwarts) {
+function displayGryffindorMembers(hogwartsMembers) {
+  // Early return if input is null, undefined, or not an array
+  if (!hogwartsMembers || !Array.isArray(hogwartsMembers)) {
+    console.error("Invalid input: Expected an array of Hogwarts members.");
+    return;
+  }
+
   // Loop through each person in the hogwarts array
-  hogwarts.forEach(({ firstName, lastName, house }) => {
+  hogwartsMembers.forEach(({ firstName, lastName, house }) => {
     // If the person belongs to Gryffindor house, log their full name
     if (house === "Gryffindor") {
       console.log(`${firstName} ${lastName}`);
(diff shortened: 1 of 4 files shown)


Sprint-1/destructuring/exercise-2/exercise.js
```diff
@@ -70,3 +70,33 @@ let hogwarts = [
     occupation: "Teacher",
   },
 ];
+
+// Task 1:
+// Function to display names of people in Gryffindor house
+function displayGryffindorMembers(hogwarts) {
```
tyzia, line 76: You are creating this function and parameters. What do you think displayGryffindorMembers() will accept as parameters? Sure, when you will invoke this method, you will pass hogwarts to it (as this variable is defined for you), but it is up to you to come up with a good name for the parameters of your function. Think what this parameter will hold as a value and name it accordingly. This is not a show-stopper for your PR but it is good to practice.
Next question, what will happen if I will do something like this:

displayGryffindorMembers(null);

tyzia (commented):
Overall looks good. Please, take a look at one comment which I left.

Sprint-1/destructuring/exercise-2/exercise.js
AFatmaa: Hi @tyzia,

Thank you for your feedback! I hadn’t considered it before, but I’ve made the improvements. I added a check at the start of each function to ensure the input is valid. If the input is null, undefined, or not an array, the function will show an error and stop. I also renamed the parameter from hogwarts to hogwartsMembers, making it clearer and more descriptive. 😊

When I called displayGryffindorMembers(null);, I got a TypeError. Inside the function, the code attempts to execute hogwarts.forEach(...), but .forEach() can only be called on an array. Since null is not an array, this throws a TypeError.

Thanks again. I’ve learned a lot! 🌸

AFatmaa:
Thank you @tyzia!  I’ve updated the code based on your suggestions. 
Let me know if there’s anything else you’d like me to improve. Thank you for helping me make my code better! 🙌

tyzia (approved):
Thank you for the updates. These are exactly what I wanted to see. Keep up good work! Approved.

AFatmaa:
Thank you @tyzia! 🤗

PR #112, opened 2025-01-01
London | Gislaine Della Bella | Module-Data-Flows | Week 2 | DEBUGGING BOOK LIBRARY

Della-Bella:
## Learners, PR Template

Self checklist

- [X ] I have committed my files one by one, on purpose, and for a reason
- [X ] I have titled my PR with COHORT_NAME | FIRST_NAME LAST_NAME | REPO_NAME | WEEK 
- [X ] I have tested my changes
- [X ] My changes follow the [style guide](https://curriculum.codeyourfuture.io/guides/contributing/)
- [X ] My changes meet the [requirements](./README.md) of this task

## Changelist

Briefly explain your PR.

## Questions

Hi volunteer,
Thank you for look at my code,
Gislaine

### DIFF
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -36,9 +36,13 @@ function submit() {
   ) {
     alert("Please fill all fields!");
     return false;
+  } else if (isNaN(pages.value) || parseInt(pages.value) < 0) {
+    // Check if pages is a non-negative number
+    alert("Please enter a valid non-negative number for pages.");
+    return false; // Prevent adding book if it is negative or not a number
   } else {
     let book = new Book(title.value, author.value, pages.value, check.checked); //passing value tittle
-    myLibrary.push(book);// syntax error (library.push)
+    myLibrary.push(book);
     render();
   }
 }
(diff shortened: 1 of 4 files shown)


halilibrahimcelik (commented):
Good Work, everything is working as expected! Here are some points that you might have to consider if you want to enhance submit form. While submitting books, if user only uses white spaces, it passes your if check.
Also you can also check input values of page field for instance. It is an unlike scenario but what if user write negative numbers. you can also prevent this within submit function
image

debugging/book-library/script.js
```diff
@@ -77,9 +77,9 @@ function render() {
     wasReadCell.appendChild(changeBut);
     let readStatus = "";
     if (myLibrary[i].check == false) {
-      readStatus = "Yes";
-    } else {
       readStatus = "No";
+    } else {
```
halilibrahimcelik, line 81: In order to increase UI&UX experience, you can also change color of "NO" btn  to red color.

Della-Bella: Hi halilibrahimcelik, 
It makes sense not add a - number. I checked with chat GTP and it suggested it for me:

1- isNaN(pages.value) check: The isNaN() function was added before the value parsing to check if the value is a valid number. / Donne
2- parseInt(pages.value) < 0 check: After checking it is a valid number, we convert pages.value to an integer using parseInt() and check if it's less than 0. If it's not a number or negative, an alert is shown, and return false will prevent the rest of submit() from being executed.

Focus on the Submit() Function function it the code works better now,

About the Red & Green book, it's all good now.  He showed me how to add a btn Dnager class with a condition.

thank you

PR #124, opened 2025-01-05
London | Emmanuel Gessessew | Module-Data-Flows| WEEK 2 Book library project

Emmanuelgessessew:
## Learners, PR Template

Self checklist

- [ yes] I have committed my files one by one, on purpose, and for a reason
- [yes ] I have titled my PR with COHORT_NAME | FIRST_NAME LAST_NAME | REPO_NAME | WEEK 
- [ yes] I have tested my changes
- [ yes] My changes follow the [style guide](https://curriculum.codeyourfuture.io/guides/contributing/)
- [ yes] My changes meet the [requirements](./README.md) of this task

## Changelist

Briefly explain your PR.
Correcting input validation before adding books.
Using myLibrary instead of library.
Fixing title/author mismatch.
Ensuring the delete button works.
Correctly saving the "read" status.

## Questions

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "liveServer.settings.port": 5502
+}
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -7,7 +7,7 @@ window.addEventListener("load", function (e) {
 
 function populateStorage() {
   if (myLibrary.length == 0) {
-    let book1 = new Book("Robison Crusoe", "Daniel Defoe", "252", true);
+    let book1 = new Book("Robinson Crusoe", "Daniel Defoe", "252", true);
     let book2 = new Book(
       "The Old Man and the Sea",
       "Ernest Hemingway",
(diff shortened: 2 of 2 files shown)


OracPrime:
It's always a good habit to try to use your code as an end user just before you submit a PR.  In particular, there's the "does it work" test.  In particular have you tried submitting books with different values of "Checked" for whether it was read or not.  Or indeed does it look right for your initial two books?  Can you see the problem here?

OracPrime:
I also when possible try to look at the diff of each changed file I submit, as this sometimes makes bugs more obvious than looking at the entire code.  For example my previous comment might be most easily addressed by looking at https://github.com/CodeYourFuture/Module-Data-Flows/pull/124/commits/7239a23ea8d1844406f2131ee97cab0977816b57 and seeing if anything leaps off the page at you!

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "liveServer.settings.port": 5502
```
OracPrime, line 2: There is debate as to whether you should commit your .vscode file to version control.  Generally it is things related to your own personal setup, and you don't want to force other users to share them.  Google .gitignore to discover how to exclude files or folders from version control.

Sometimes you do want local configuration to be shared, but only do it intentionally!

debugging/book-library/script.js
```diff
@@ -58,5 +58,17 @@ function render() {
...
+  let row = table.insertRow(1);
+  let titleCell = row.insertCell(0);
+  let authorCell = row.insertCell(1);
+  let pagesCell = row.insertCell(2);
+  let wasReadCell = row.insertCell(3);
+  let deleteCell = row.insertCell(4);
 
+  titleCell.innerHTML = myLibrary[i].title;
+  authorCell.innerHTML = myLibrary[i].author;
+  pagesCell.innerHTML = myLibrary[i].pages;
```
OracPrime, line 73: I'm pretty sure your code wasn't working when you made this commit, because there are a whole bunch of missing } at the end (which show up in the next commit).  It's generally considered prudent to have committed code in a reasonable state of "cleanness" as others may check it out and try to use it.  Less of an issue on a branch that only you are using, but something to bear in mind

debugging/book-library/script.js
```diff
@@ -53,51 +54,35 @@ function Book(title, author, pages, check) {
...
-    let pagesCell = row.insertCell(2);
-    let wasReadCell = row.insertCell(3);
-    let deleteCell = row.insertCell(4);
-    titleCell.innerHTML = myLibrary[i].title;
-    authorCell.innerHTML = myLibrary[i].author;
-    pagesCell.innerHTML = myLibrary[i].pages;
+// Insert updated rows and cells
+let length = myLibrary.length;
+for (let i = 0; i < length; i++) {
+  let row = table.insertRow(1);
```
OracPrime, line 64: What does the 1 argument do here?  How does the result differ if you leave it out? Which do you prefer?

debugging/book-library/script.js
```diff
@@ -7,7 +7,7 @@ window.addEventListener("load", function (e) {
 
 function populateStorage() {
   if (myLibrary.length == 0) {
-    let book1 = new Book("Robison Crusoe", "Daniel Defoe", "252", true);
+    let book1 = new Book("Robinson Crusoe", "Daniel Defoe", "252", true);
```
OracPrime, line 10: This is a good correction.  Having accurate test data in software helps users and testers take it more seriously.  So well done for bothering to correct this.

debugging/book-library/script.js
```diff
@@ -53,51 +54,35 @@ function Book(title, author, pages, check) {
...
+  authorCell.innerHTML = myLibrary[i].author;
+  pagesCell.innerHTML = myLibrary[i].pages;
 
-    changeBut.addEventListener("click", function () {
-      myLibrary[i].check = !myLibrary[i].check;
-      render();
-    });
+// Add a delete button to every row and render again
+let delButton = document.createElement("button");
+delButton.id = i + 5;
```
OracPrime, line 77: What is this delButton.id = i+5 doing?

debugging/book-library/script.js
```diff
@@ -53,51 +54,35 @@ function Book(title, author, pages, check) {
...
-    delBut.className = "btn btn-warning";
-    delBut.innerHTML = "Delete";
-    delBut.addEventListener("clicks", function () {
-      alert(`You've deleted title: ${myLibrary[i].title}`);
-      myLibrary.splice(i, 1);
-      render();
-    });
-  }
+delButton.addEventListener("click", function () {
+  alert(`You've deleted title: ${myLibrary[i].title}`);
```
OracPrime, line 83: Explain to me where the value for i comes from here?

debugging/book-library/script.js
```diff
@@ -53,51 +54,35 @@ function Book(title, author, pages, check) {
...
+  myLibrary.splice(i, 1);
+  render();
+});
```
OracPrime, line 86: If you can set a formatter such as prettier up on your vscode, you'll be able to easily format the source file so that the structure is clearer.

OracPrime (commented):
Very close, but a few areas where improvement and/or increased understanding is possible

debugging/book-library/script.js
Emmanuelgessessew: The 1 argument makes the row insert as the second row, below the header. Without it, the row is added at the end of the table. I prefer using 1 to keep new rows consistently under the header.

debugging/book-library/script.js
Emmanuelgessessew: thanks

debugging/book-library/script.js
Emmanuelgessessew: delButton.id = i + 5; gives each delete button a unique id by starting at 5 and incrementing with each iteration. This avoids conflicts with other IDs.

debugging/book-library/script.js
Emmanuelgessessew: The value of i comes from the loop, but using let ensures the correct value is captured for each iteration. Without it, i might not reference the correct index.

debugging/book-library/script.js
Emmanuelgessessew: Thanks for the suggestion! I'll set up Prettier in VS Code to keep the code well-structured and easier to read

debugging/book-library/script.js
OracPrime: If that's what you want, that's fine.  It does of course reverse the order of the items...

debugging/book-library/script.js
OracPrime: Does it need an id at all?

PR #131, opened 2025-01-09
West Midlands | Gabriel Deng | Module-Data-Groups | Week 2 - Book reading

gai93003:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -76,7 +76,7 @@ function render() {
     changeBut.className = "btn btn-success";
     wasReadCell.appendChild(changeBut);
     let readStatus = "";
-    if (myLibrary[i].check == false) {
+    if (myLibrary[i].check) {
       readStatus = "Yes";
     } else {
       readStatus = "No";
@@ -90,11 +90,11 @@ function render() {
 
     //add delete button to every row and render again
     let delButton = document.createElement("button");
-    delBut.id = i + 5;
-    deleteCell.appendChild(delBut);
-    delBut.className = "btn btn-warning";
-    delBut.innerHTML = "Delete";
-    delBut.addEventListener("clicks", function () {
+    delButton.id = i + 5;
+    deleteCell.appendChild(delButton);
+    delButton.className = "btn btn-warning";
+    delButton.innerHTML = "Delete";
+    delButton.addEventListener("clicks", function () {
       alert(`You've deleted title: ${myLibrary[i].title}`);
       myLibrary.splice(i, 1);
       render();
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -38,7 +38,7 @@ function submit() {
     return false;
   } else {
     let book = new Book(title.value, author.value, pages.value, check.checked);
-    myLibrarylibrary.push(book);
+    myLibrary.push(book);
     render();
   }
 }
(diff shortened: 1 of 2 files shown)


debugging/code-reading/readme.md
```diff
@@ -35,6 +38,8 @@ console.log(y);
 
 What will be the output of this code. Explain your answer in 50 words or less.
 
+Ans: The output will be 10 and undefined. This is because the y variable is declared within the scope of the function and therefore cannot be globally accessed.
```
cjyuan, line 41: Did you try running the code to see what it produces?

The second statement does not explain why the first output is 10.

debugging/code-reading/readme.md
```diff
@@ -17,6 +17,9 @@ Take a look at the following code:
 
 Explain why line 5 and line 8 output different numbers.
 
+Ans: The variable at line 5  is a globally declared as such it will return 1, while the variable at line 8 is a scope variable since is declared with the scope of the function.
+      Therefore, when the function is called it will console the x variable declared in its scope.
+
```
cjyuan, line 22: I am not sure what you meant. The `x` at line 5 is certainly not globally declared.
I wonder how ChatGPT would answer this question.

cjyuan (commented):
In terms of input validation, 
1. Can `.value` be `null`?

debugging/book-library/script.js
```diff
@@ -54,7 +54,7 @@ function render() {
   let table = document.getElementById("display");
   let rowsNumber = table.rows.length;
   //delete old table
-  for (let n = rowsNumber - 1; n > 0; n-- {
+  for (let n = rowsNumber - 1; n > 0; n--) {
```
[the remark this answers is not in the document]
gai93003: I think the most efficient way is to set the innerHTML of the table to emtpy and then appendChild header. In this case the first row.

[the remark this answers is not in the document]
gai93003: I have also implemented the checking of the input of spaces using the input.value.trim() and check if it is empty. If so, an alert will pop out. Similarly, the validation of the page input field is also implemented.

debugging/book-library/script.js
```diff
@@ -90,11 +90,11 @@ function render() {
 
     //add delete button to every row and render again
     let delButton = document.createElement("button");
-    delBut.id = i + 5;
-    deleteCell.appendChild(delBut);
-    delBut.className = "btn btn-warning";
-    delBut.innerHTML = "Delete";
-    delBut.addEventListener("clicks", function () {
+    delButton.id = i + 5;
```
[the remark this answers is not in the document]
gai93003: The assignment of the id in this case is not important because there is no where in the code where the button is being used in reference to the id. So i have deleted it to make the code more readable and as precise as possible.

debugging/code-reading/readme.md
```diff
@@ -62,3 +67,8 @@ console.log(y);
 ```
 
 What will be the output of this code. Explain your answer in 50 words or less.
+
+Ans: The output will be 9 and {x: 10}; This is due to the difference by which each of the
```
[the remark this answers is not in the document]
gai93003: Thanks

debugging/code-reading/readme.md
gai93003: You are right, the output is indeed 10 and undefined. I try making some research and it turns out that it is because the function has no return value although y has been declared and assigned a value.

debugging/code-reading/readme.md
gai93003: I misinterpreted how the scope works. I didn't take into account where the console logs where being called. The console inside the function will refer to the local variable x and vice-versa.

debugging/code-reading/readme.md
cjyuan: When I ran the script, the script threw an exception (because `y` is not declared). Just curious, did you not get an exception?

PR #134, opened 2025-01-10
West Midlands| Danial Bashirzadeh | Module-Data-Flows | Week2 | Debugging | book-library

bashirzadehdanial:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -37,8 +37,8 @@ function submit() {
     alert("Please fill all fields!");
     return false;
   } else {
-    let book = new Book(title.value, title.value, pages.value, check.checked);
-    library.push(book);
+    let book = new Book(title.value, author.value, pages.value, check.checked);
+    myLibrary.push(book);
     render();
   }
 }
@@ -54,7 +54,7 @@ function render() {
   let table = document.getElementById("display");
   let rowsNumber = table.rows.length;
   //delete old table
-  for (let n = rowsNumber - 1; n > 0; n-- {
+  for (let n = rowsNumber - 1; n > 0; n--) {
     table.deleteRow(n);
   }
   //insert updated row and cells
@@ -76,7 +76,7 @@ function render() {
     changeBut.className = "btn btn-success";
     wasReadCell.appendChild(changeBut);
     let readStatus = "";
-    if (myLibrary[i].check == false) {
+    if (myLibrary[i].check) {
       readStatus = "Yes";
     } else {
       readStatus = "No";
@@ -90,11 +90,11 @@ function render() {
 
     //add delete button to every row and render again
     let delButton = document.createElement("button");
-    delBut.id = i + 5;
-    deleteCell.appendChild(delBut);
-    delBut.className = "btn btn-warning";
-    delBut.innerHTML = "Delete";
-    delBut.addEventListener("clicks", function () {
+    delButton.id = i + 5;
+    deleteCell.appendChild(delButton);
+    delButton.className = "btn btn-warning";
+    delButton.innerHTML = "Delete";
+    delButton.addEventListener("click", function () {
       alert(`You've deleted title: ${myLibrary[i].title}`);
       myLibrary.splice(i, 1);
       render();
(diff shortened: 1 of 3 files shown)


debugging/book-library/script.js
```diff
@@ -54,7 +54,7 @@ function render() {
   let table = document.getElementById("display");
   let rowsNumber = table.rows.length;
   //delete old table
-  for (let n = rowsNumber - 1; n > 0; n-- {
+  for (let n = rowsNumber - 1; n > 0; n--) {
```
cjyuan, line 57: Can you think of a more efficient way to remove all rows (except the heading) in the table?

debugging/book-library/script.js
```diff
@@ -90,11 +90,11 @@ function render() {
 
     //add delete button to every row and render again
     let delButton = document.createElement("button");
-    delBut.id = i + 5;
-    deleteCell.appendChild(delBut);
-    delBut.className = "btn btn-warning";
-    delBut.innerHTML = "Delete";
-    delBut.addEventListener("clicks", function () {
+    delButton.id = i + 5;
```
cjyuan, line 93: - Do you think the value assigned to this `id` attribute is unique?
- Do you think there is a need to assign any id attribute to `delButton`?
- Do you think there is a need to assign any id attribute to `changeBut` (at line 75)?

debugging/code-reading/readme.md
```diff
@@ -17,6 +17,8 @@ Take a look at the following code:
 
 Explain why line 5 and line 8 output different numbers.
 
+The variable at line 5 is global, so it returns 1, while the variable at line 8 is local to the function. Thus, when the function is called, it logs the x variable from its scope.
```
cjyuan, line 20: The variable at line 5 is definite not global. 
I wonder how ChatGPT would answer this question.

debugging/code-reading/readme.md
```diff
@@ -35,6 +37,8 @@ console.log(y);
 
 What will be the output of this code. Explain your answer in 50 words or less.
 
+The output will be 10 and undefined, as the y variable is scoped within the function and can't be accessed globally.
```
cjyuan, line 40: Did you try executing the code to verify your answer?

The second statement is correct about `y`, but it does not explain why the first output is 10.

debugging/code-reading/readme.md
```diff
@@ -62,3 +66,5 @@ console.log(y);
 ```
 
 What will be the output of this code. Explain your answer in 50 words or less.
+
+The output will be 9 and {x: 10} because of the different ways the variables are passed. In f1, x is passed by value, so the function doesn't alter its original value. In f2, y is passed by reference, so changes within the function modify the original value of y.
```
cjyuan, line 70: Spot on.

cjyuan (commented):
In terms of input validation, 
1. Can `.value` be `null`?

debugging/book-library/script.js
bashirzadehdanial: A more efficient way to remove all rows except the heading is to clear the tbody directly.

debugging/book-library/script.js
bashirzadehdanial: 1- No, the id values aren't guaranteed to be unique.
2- No, id for delButton is unnecessary.
3- No, id for changeBut isn't needed.
4- More descriptive names like deleteButton and toggleReadButton.

debugging/code-reading/readme.md
bashirzadehdanial: I provided the code to chatGPT and received "Line 5 logs 2, Line 8 logs 1", which I believe is wrong as the output logs 2 for both because the variable is updated.

bashirzadehdanial: I meant line 1 is global, not line 5, it was my bad.

debugging/code-reading/readme.md
bashirzadehdanial: Output is 10 because it accesses the global variable x.

PR #142, opened 2025-01-13
West Midlands | Samira Hekmati | Module-Data-Flows | Sprint-2 | Book-library

samirahekmati:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -1,6 +1,6 @@
 let myLibrary = [];
 
-window.addEventListener("load", function (e) {
+window.addEventListener("load", function () {
   populateStorage();
   render();
 });
@@ -31,14 +31,18 @@ function submit() {
   if (
     title.value == null ||
     title.value == "" ||
+    author.value == null ||
+    author.value == "" ||
     pages.value == null ||
     pages.value == ""
   ) {
     alert("Please fill all fields!");
     return false;
   } else {
-    let book = new Book(title.value, title.value, pages.value, check.checked);
-    library.push(book);
+    //pass author value correctly
+    let book = new Book(title.value, author.value, pages.value, check.checked);
+    //Replace library.push(book) with myLibrary.push(book)
+    myLibrary.push(book);
     render();
   }
 }
@@ -54,7 +58,7 @@ function render() {
   let table = document.getElementById("display");
   let rowsNumber = table.rows.length;
   //delete old table
-  for (let n = rowsNumber - 1; n > 0; n-- {
+  for (let n = rowsNumber - 1; n > 0; n--) {
     table.deleteRow(n);
   }
   //insert updated row and cells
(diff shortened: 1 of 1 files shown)


debugging/book-library/script.js
```diff
@@ -90,11 +89,11 @@ function render() {
 
     //add delete button to every row and render again
     let delButton = document.createElement("button");
-    delBut.id = i + 5;
-    deleteCell.appendChild(delBut);
-    delBut.className = "btn btn-warning";
-    delBut.innerHTML = "Delete";
-    delBut.addEventListener("clicks", function () {
+    delButton.id = i + 5;
```
cjyuan, line 92: - Do you think the value assigned to this `id` attribute is unique?
- Do you think there is a need to assign any id attribute to `delButton`?
- Do you think there is a need to assign any id attribute to `changeBut` (at line 79)?

debugging/book-library/script.js
```diff
@@ -54,7 +58,7 @@ function render() {
   let table = document.getElementById("display");
   let rowsNumber = table.rows.length;
   //delete old table
-  for (let n = rowsNumber - 1; n > 0; n-- {
+  for (let n = rowsNumber - 1; n > 0; n--) {
```
cjyuan, line 61: Can you think of a more efficient way to remove all rows (except the `<th>...</th>`) in the table?

cjyuan (commented):
In terms of input validation, 

debugging/book-library/script.js
samirahekmati: you're right @cjyuan The value i + 5 for the id attribute isn't guaranteed to be unique.
Instead of using i + 5 (which is not reliable),I can simply use the i value as the id for the delete button since it's guaranteed to be unique within the context of the current render.

samirahekmati: infact no need to assign an ID here. I am already using the i index inside the click event listener to know which book was clicked (since i uniquely identifies each book in the loop).
I can simply remove the assignment of the id for the delete button.

samirahekmati: Similarly to the delButton, there is no need to assign an id attribute to the changeBut button (line 79).

samirahekmati: changeBut → toggleReadButton: The button's purpose is to toggle the "read" status of the book, so a more descriptive name would help clarify its role.
delButton → deleteButton: This name is already clear but can be made consistent with the other button name by making it follow the same pattern.

samirahekmati:
Can .value be null? (Do we need to check someInputElement.value == null?)
Answer: No, .value will not be null on a valid input element. The .value of an input field is an empty string ("") if the user has not entered anything, not null. So, checking for value == null is unnecessary.

debugging/book-library/script.js
samirahekmati: Yes!  simply clear all the rows except the header row by removing all child elements from the <tbody> (or equivalent body section of the table). This avoids looping through the rows and directly empties the content of the table

PR #143, opened 2025-01-13
West-Midlands-Community | Matthew Law | Module-Data-Flows | WEEK 11 | Book Library

matthewlawpixel:
Self checklist

## Changelist

PR is for completing exercises.

## Questions

### DIFF
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -1,22 +1,15 @@
 let myLibrary = [];
 
-window.addEventListener("load", function (e) {
+window.addEventListener("load", function () {
   populateStorage();
   render();
 });
 
 function populateStorage() {
-  if (myLibrary.length == 0) {
-    let book1 = new Book("Robison Crusoe", "Daniel Defoe", "252", true);
-    let book2 = new Book(
-      "The Old Man and the Sea",
-      "Ernest Hemingway",
-      "127",
-      true
-    );
-    myLibrary.push(book1);
-    myLibrary.push(book2);
-    render();
+  if (myLibrary.length === 0) {
+    let book1 = new Book("Robinson Crusoe", "Daniel Defoe", "252", true);
+    let book2 = new Book("The Old Man and the Sea", "Ernest Hemingway", "127", true);
+    myLibrary.push(book1, book2);
   }
 }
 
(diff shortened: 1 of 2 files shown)


debugging/book-library/script.js
```diff
@@ -53,51 +43,41 @@ function Book(title, author, pages, check) {
 function render() {
   let table = document.getElementById("display");
   let rowsNumber = table.rows.length;
-  //delete old table
-  for (let n = rowsNumber - 1; n > 0; n-- {
+
+  // Delete old rows
+  for (let n = rowsNumber - 1; n > 0; n--) {
```
cjyuan, line 48: Can you think of a more efficient way to remove all rows (except the `<th>...</th>`) in the table?

debugging/book-library/script.js
```diff
@@ -53,51 +43,41 @@ function Book(title, author, pages, check) {
...
 
-    //add delete button to every row and render again
+    let deleteCell = row.insertCell(4);
     let delButton = document.createElement("button");
-    delBut.id = i + 5;
-    deleteCell.appendChild(delBut);
-    delBut.className = "btn btn-warning";
-    delBut.innerHTML = "Delete";
-    delBut.addEventListener("clicks", function () {
+    delButton.id = i + 5;
```
cjyuan, line 73: - Is the value assigned to this `id` attribute unique?
- Is there a need to assign an id attribute to `delButton`?
- Is there a need to assign an id attribute to `changeBut` (at line 62)?

debugging/book-library/script.js
```diff
@@ -53,51 +43,41 @@ function Book(title, author, pages, check) {
...
+    delButton.className = "btn btn-warning";
+    delButton.innerHTML = "Delete";
+    delButton.addEventListener("click", function () {
       alert(`You've deleted title: ${myLibrary[i].title}`);
       myLibrary.splice(i, 1);
       render();
     });
```
cjyuan, line 80: At the moment when the alert message is displayed, has the book mentioned in the message been deleted yet?

cjyuan (commented):
All the bugs described in the spec have been successfully fixed.

debugging/book-library/script.js
matthewlawpixel: Yes, there is a more efficient way to remove all rows (except the <th> elements) in the table. The entire <tbody> can be cleared in a single operation by setting its innerHTML to an empty string.

debugging/book-library/script.js
matthewlawpixel: When the alert message is displayed, the book mentioned in the message has not been deleted yet. I have made changes to the code so that the book is delated before the alert is displayed.

matthewlawpixel: Thank you for pointing this out @cjyuan. I have moved the alert message to occur after the book deletion

debugging/book-library/script.js
matthewlawpixel: @cjyuan Yes you are right. After reviewing the need for id attributes, I determined that they are not necessary I’ve opted to use data-* attributes to manage element associations. I have also changed the variable name for the buttons to be more clear "deleteButton" and "toggleButton"

matthewlawpixel:
Hi @cjyuan if everything is alright here. I'll make this as complete. Thank you for helping me reviewing my code.

cjyuan:
yes. all good

PR #145, opened 2025-01-15
WM4 | FATIMA SAFANA | Module-Data-Flows | Book library 

Fatimasfn:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -38,7 +38,7 @@ function submit() {
     alert("Please fill all fields!");
     return false;
   } else {
-    let book = new Book(title.value, title.value, pages.value, check.checked);
+    let book = new Book(title.value, author.value, pages.value, check.checked);
     myLibrary.push(book);
     render();
   }
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -17,7 +17,7 @@ function populateStorage() {
     );
     myLibrary.push(book1);
     myLibrary.push(book2);
-    render();
+    //render();
   }
 }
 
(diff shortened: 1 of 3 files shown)


debugging/book-library/script.js
```diff
@@ -89,12 +97,12 @@ function render() {
...
 
     //add delete button to every row and render again
-    let delButton = document.createElement("button");
+    let delBut = document.createElement("button");
     delBut.id = i + 5;
     deleteCell.appendChild(delBut);
     delBut.className = "btn btn-warning";
     delBut.innerHTML = "Delete";
-    delBut.addEventListener("clicks", function () {
+    delBut.addEventListener("click", function () {
```
cjyuan, line 105: - At line 101, is the value assigned to this `id` attribute unique?
- Is there a need to assign an id attribute to `delBut` (at line 101)?
- Is there a need to assign an id attribute to `changeBut` (at line 83)?

debugging/book-library/script.js
```diff
@@ -89,12 +97,12 @@ function render() {
...
       alert(`You've deleted title: ${myLibrary[i].title}`);
       myLibrary.splice(i, 1);
       render();
```
cjyuan, line 108: At the moment when the alert message is displayed, has the book mentioned in the message been deleted yet?
How would you rearrange these statements to make the message truthful?

cjyuan (commented):
All the bugs explicitly described in the spec have been fixed.

In terms of input validation, 

debugging/book-library/script.js
Fatimasfn: no, I think its not needed

PR #151, opened 2025-01-19
London11 | Karla Grajales | Module-Data-Groups | Sprint-3 Debugging    Book-Library-Book

Grajales-K:
## Learners, PR Template

Self checklist

- [X] I have titled my PR with COHORT_NAME | FIRST_NAME LAST_NAME | REPO_NAME | WEEK 
- [X] My changes follow the [style guide](https://curriculum.codeyourfuture.io/guides/contributing/)

## Changelist

This project taught us key skills in debugging and version control. I resolved Git branch issues, handled cherry-picks, and fixed event listener errors in JavaScript. Clear boolean condition checks improved code reliability, and we efficiently resolved merge conflicts during cherry-pick operations. Finally, I emphasized the importance of maintaining a disciplined Git workflow to manage changes systematically.

## Questions

### DIFF
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -76,7 +76,7 @@ function render() {
     changeBut.className = "btn btn-success";
     wasReadCell.appendChild(changeBut);
     let readStatus = "";
-    if (myLibrary[i].check == "false") {
+    if (myLibrary[i].check == false) {
       readStatus = "No";
     } else {
       readStatus = "Yes";
@@ -88,8 +88,7 @@ function render() {
       render();
     });
 
-
-    // second option to booleans comparisons
+    // // second option to booleans comparisons
     // let changeBut = document.createElement("button");
     // changeBut.id = i;
     // changeBut.className = "btn btn-success";
@@ -110,16 +109,15 @@ function render() {
     //   render();
     // });
 
-
     //add delete button to every row and render again
     const delButton = document.createElement("button");
     delButton.id = i + 5;
     deleteCell.appendChild(delButton);
     delButton.className = "btn btn-warning";
     delButton.innerHTML = "Delete";
-    delButton.addEventListener("click", function (index) {
+    delButton.addEventListener("click", function () {
       alert(`You've deleted title: ${myLibrary[i].title}`);
-      myLibrary.splice(index, 1);
+      myLibrary.splice(i, 1);
       render();
     });
   }
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -76,10 +76,10 @@ function render() {
     changeBut.className = "btn btn-success";
     wasReadCell.appendChild(changeBut);
     let readStatus = "";
-    if (myLibrary[i].check == false) {
-      readStatus = "Yes";
-    } else {
+    if (myLibrary[i].check == "false") {
       readStatus = "No";
+    } else {
+      readStatus = "Yes";
     }
     changeBut.innerText = readStatus;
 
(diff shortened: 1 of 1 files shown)


debugging/book-library/script.js
```diff
@@ -54,7 +54,7 @@ function render() {
   let table = document.getElementById("display");
   let rowsNumber = table.rows.length;
   //delete old table
-  for (let n = rowsNumber - 1; n > 0; n-- {
+  for (let n = rowsNumber - 1; n > 0; n--) {
```
cjyuan, line 57: Question:
Can you think of a more efficient way to remove all rows (except the `<th>...</th>`) in the table?

debugging/book-library/script.js
```diff
@@ -88,13 +88,34 @@ function render() {
...
+
     //add delete button to every row and render again
-    let delButton = document.createElement("button");
-    delBut.id = i + 5;
-    deleteCell.appendChild(delBut);
-    delBut.className = "btn btn-warning";
-    delBut.innerHTML = "Delete";
-    delBut.addEventListener("clicks", function () {
+    const delButton = document.createElement("button");
+    delButton.id = i + 5;
```
cjyuan, line 114: - Is the value assigned to this `id` attribute unique (line 114)?
- Is there a need to assign an id attribute to `delButton` (line 114)?
- Is there a need to assign an id attribute to `changeBut` (at line 75)?

debugging/book-library/script.js
```diff
@@ -88,13 +88,34 @@ function render() {
...
+    deleteCell.appendChild(delButton);
+    delButton.className = "btn btn-warning";
+    delButton.innerHTML = "Delete";
+    delButton.addEventListener("click", function () {
       alert(`You've deleted title: ${myLibrary[i].title}`);
       myLibrary.splice(i, 1);
       render();
```
cjyuan, line 121: At the moment when the alert message is displayed, has the book mentioned in the message been deleted yet?
How would you rearrange these statements to make the message truthful?

cjyuan (commented):
- On page load, is it necessary to call `render()` twice?

- In terms of input validation, 

There is also some errors in `index.html`. Perhaps an HTML validator can quickly help you identify them.

debugging/book-library/script.js
Grajales-K: > * Is the value assigned to this `id` attribute unique (line 114)?
> * Is there a need to assign an id attribute to `delButton` (line 114)?
> * Is there a need to assign an id attribute to `changeBut` (at line 75)?
> * Can you think of a more consistent way to name the variables representing the two buttons?

1 & 2. we are not using Id, I deleted
3 & 4. Yes, I change the name for better readability.

debugging/book-library/script.js
Grajales-K: Hi @cjyuan, I changed the alert message.

Grajales-K:
> * On page load, is it necessary to call `render()` twice?
> * In terms of input validation,
> 
> 1. Are all input properly checked?
> 2. Can `.value` be `null`? (Do we need to check `someInputElement.value == null`?)
> 3. What if a user enters only space characters in the "title" input field?
> 4. What if a users enters `-1` or `3.1416` in the "pages" input field?
> 
> There is also some errors in `index.html`. Perhaps an HTML validator can quickly help you identify them.

debugging/book-library/script.js
Grajales-K: Can you please give more detail of this question because I don't understand what you expect? thanks

cjyuan: Clearing the table row by row is relatively slow. If we can just clear one of the elements in the HTML to accomplish the same, then doing so is preferred. (But for an application this simple, performance is not an issue)

debugging/book-library/script.js
cjyuan: This message seems worse. :D What if the user is "unsure" about carrying out the action and wants to cancel the action?
`alert(`You are deleting: "${myLibrary[i].title}" Are you sure?!`);`

What I was suggesting was, you can perform the deletion first, update the view, then prompt the message.

PR #163, opened 2025-03-20
LONDON-JAN-25 | ANDREI FILIPPOV | Module-Data Flows | WEEK 2 Feature/book-library

Droid-An:
## Learners, PR Template

Self checklist

## Changelist

fixed bug in book library app

## Questions

no questions

### DIFF
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -68,18 +68,18 @@ function render() {
   //insert updated row and cells
   const length = myLibrary.length;
   for (let i = 0; i < length; i++) {
-    let row = table.insertRow(1);
-    let titleCell = row.insertCell(0);
-    let authorCell = row.insertCell(1);
-    let pagesCell = row.insertCell(2);
-    let wasReadCell = row.insertCell(3);
-    let deleteCell = row.insertCell(4);
+    const row = table.insertRow(1);
+    const titleCell = row.insertCell(0);
+    const authorCell = row.insertCell(1);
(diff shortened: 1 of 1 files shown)


jenny-alexander (changes requested):
You fixed all of the problems outlined in the requirements. It works well now. However, there are a few more changes you implement to make your code even better:

1. The browser tab currently shows '127.0.0.1:555/debugging/book-library. Can you update it to be more explanatory to the user. If you want to read more about this here is a doc: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/title

2. Can you review all of the places that you declared a variable with 'let'? In general, you will mostly always use const (and this should be your default). If you need to reassign a value to the variable, then you would use let. Here is a video you can watch: https://youtu.be/RE6qf3As-XU?si=_YaiEPeBC5Cx_j56

3. There are a few places where the variable name was not clear to me. For example, using 'but' to denote a button might be confusing to someone reviewing your code. Can you replace with a better name where you used 'but'? 🙂.

4. Should the code-reading/readme.md file be included in this PR?

Droid-An:
Thank you for the comments. 
1. I’ve updated the tab title to "Book Library App".
2. Understood — I replaced let with const wherever appropriate.
3. I also gave the variables more self-explanatory names.
4. Regarding the Code Reading exercise in the readMe file — I believe it’s optional. The issue for this task says:
> There's a Code Reading exercise included in the debugging folder that will help you think through this project.

So I completed the exercise to help improve my understanding.

PR #173, opened 2025-03-26
ITP-JAN-25 | London | Saba Farjamfard | Module-Data-Flows | Book Library App | Week-2

sabafarjamfard:
## Learners, PR Template

Self checklist

## Changelist

- Implemented dynamic book list rendering from a form
- Added support for toggling 'read' status and deleting books
- Styled the form and table with Bootstrap
- Added two initial book entries on page load
- Replaced unnecessary `let` declarations with `const`
- Updated the `<title>` tag to "My Book Library"

Briefly explain your PR.

## Questions

### DIFF
--- a/debugging/book-library/index.html
+++ b/debugging/book-library/index.html
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <title> </title>
+    <title>My Book LibraryFiu </title>
     <meta
       charset="utf-8"
       name="viewport"
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -1,26 +1,23 @@
-let myLibrary = [];
+const myLibrary = [];
 
-window.addEventListener("load", function (e) {
+window.addEventListener("load", function () {
   populateStorage();
 });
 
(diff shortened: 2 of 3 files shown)


jenny-alexander (changes requested):
You fixed all of the problems outlined in the requirements. It works well now. However, there are a few more changes you implement to make your code even better:

1. The browser tab currently shows '127.0.0.1:555/debugging/book-library. Can you update it to be more explanatory to the user. If you want to read more about this here is a doc: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/title
2. Can you review all of the places that you declared a variable with 'let'? In general, you will mostly always use const (and this should be your default). If you need to reassign a value to the variable, then you would use let. Here is a video you can watch: https://youtu.be/RE6qf3As-XU?si=_YaiEPeBC5Cx_j56
3. The changelist portion of your pull request is currently empty. This is an important part that helps other engineers understand what you've modified. Could you please add a brief description of your changes in the changelist section? A simple summary is good enough!

sabafarjamfard:
Thank you for the review. I did update it.

PR #174, opened 2025-03-27
ITP-JAN|LONDON|FOROGH AGHAEIYARIJANI|BOOK LIBRARY PROJECT|WEE 2|problem solution

Forogh-Aghaeiyarijani:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -1,23 +1,18 @@
 let myLibrary = [];
 
-window.addEventListener("load", function (e) {
+window.addEventListener("load", function () {
   populateStorage();
   render();
 });
 
 function populateStorage() {
-  if (myLibrary.length == 0) {
-    let book1 = new Book("Robison Crusoe", "Daniel Defoe", "252", true);
-    let book2 = new Book(
-      "The Old Man and the Sea",
-      "Ernest Hemingway",
-      "127",
-      true
-    );
+  if (myLibrary.length === 0) {
+    let book1 = new Book("Robinson Crusoe", "Daniel Defoe", "252", true);
+    let book2 = new Book("The Old Man and the Sea", "Ernest Hemingway", "127", true);
     myLibrary.push(book1);
     myLibrary.push(book2);
-    render();
   }
+  render();
 }
 
 const title = document.getElementById("title");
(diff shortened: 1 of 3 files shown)


jenny-alexander (changes requested):
The fixes you implemented work well and it now meets all of the requirements of this task. Here are a few things you should implement in order to make your code even better!
- Can you review all of the places that you declared a variable with 'let'? In general, you will mostly always use `const` (and this should be your default). If you need to reassign a value to the variable, then you would use let. Here is a video you can watch: https://youtu.be/RE6qf3As-XU?si=_YaiEPeBC5Cx_j56
- The changelist portion of your pull request is currently empty. This is an important part that helps other engineers understand what you've modified. Could you please add a brief description of your changes in the changelist section? A simple summary is good enough!
- Can you check off the list of requirements within the 'Self checklist' part of the PR? This is a good way to review what you need to do before submitting your PR. :)

⭐ This is a recommendation (you don't have to make this change but I highly recommend it). If you look within the package.json file, you'll see that there is a script named `format`. If you run this script from the terminal, it will format your code to have a consistent style. The file `.prettierrc` defines the formatting that will done if you run this script.

Forogh-Aghaeiyarijani:
Thank you so much for reviewing my code.I did it

debugging/book-library/script.js
```diff
@@ -1,45 +1,34 @@
 let myLibrary = [];
```
jenny-alexander, line 1: This is an example of a variable that should be declared as a const. 
There are other variables within script.js that are declared as let but should be const.  
➡️ Can you find them and change them? 
_Using const prevents accidental reassignment of the variable and makes your code easier to read._

jenny-alexander (changes requested):
Most of the changes were implemented. Just missing this one:
Can you explain your changes within the 'Change List' part of the PR?
Screenshot 2025-04-05 at 11 12 42 PM

Forogh-Aghaeiyarijani:
Changelist:

Added logic to display a table of books with ability to add/remove/update read status.

Cleaned up variable declarations by replacing let with const where appropriate.

Ensured form resets on successful book submission.

debugging/book-library/script.js
jenny-alexander, line 1: There are 2 other let variables that should be changed to const since they are not being reassigned. Can you find them and make the change please?

jenny-alexander (changes requested):
The change list portion still needs to be updated. Please update _directly_ within the **changelist** section!

debugging/book-library/script.js
Forogh-Aghaeiyarijani: I did it.

jenny-alexander: Still one more left. :)

Forogh-Aghaeiyarijani: I thought I should use Let for array. Thanks, that's it.

PR #183, opened 2025-04-05
ITP_LONDON | PRISCILLA_EMEBO | Module-Data-Flows | Book library

Priscilla-EM:
## Learners, PR Template

Self checklist

## Changelist
This PR implements a simple Book Library web app using HTML, CSS, and JavaScript. Users can add books with a title, author, page count, and read status. Each book is displayed in a table, with options to toggle its read status or delete it

## Questions

### DIFF
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -31,13 +31,15 @@ function submit() {
   if (
     title.value == null ||
     title.value == "" ||
+    author.value == null ||
+    author.value == "" ||
     pages.value == null ||
     pages.value == ""
   ) {
     alert("Please fill all fields!");
     return false;
   } else {
-    let book = new Book(title.value, title.value, pages.value, check.checked);
+    let book = new Book(title.value, author.value, pages.value, check.checked);
     library.push(book);
     render();
   }
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -54,7 +54,7 @@ function render() {
   let table = document.getElementById("display");
   let rowsNumber = table.rows.length;
   //delete old table
-  for (let n = rowsNumber - 1; n > 0; n-- {
+  for (let n = rowsNumber - 1; n > 0; n-- ){
     table.deleteRow(n);
   }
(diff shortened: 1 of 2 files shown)


debugging/book-library/script.js
```diff
@@ -31,14 +31,16 @@ function submit() {
   if (
     title.value == null ||
     title.value == "" ||
+    author.value == null ||
+    author.value == "" ||
     pages.value == null ||
     pages.value == ""
```
cjyuan, line 37: In terms of input validation, 

debugging/book-library/script.js
```diff
@@ -54,7 +56,7 @@ function render() {
   let table = document.getElementById("display");
   let rowsNumber = table.rows.length;
   //delete old table
-  for (let n = rowsNumber - 1; n > 0; n-- {
+  for (let n = rowsNumber - 1; n > 0; n-- ){
     table.deleteRow(n);
   }
```
cjyuan, line 61: Instead of deleting the table rows one by one, can you think of a more efficient way to remove all rows (except the `<th>...</th>`) in the table?

debugging/book-library/script.js
cjyuan, line 1: Line 71: With the way the book's title is assigned to an HTML element, a book with a title containing special character sequence such as `<i>` can possibly ruin the display.

Lines 77, 95: 
- Is the value assigned to these `id` attribute unique?
- Is there a need to assign an id attribute to `delButton`?
- Is there a need to assign an id attribute to `changeBut` (at line 72)?

Lines 76, 94:
- Can you think of a more consistent way to name the variables representing the two buttons, `changeBut` and `delButton`?

cjyuan (commented):
I think there are some improvements you can still make to the JS code.

Also, when I checked the code in `index.html` at https://validator.w3.org/, it indicated some errors.

cjyuan:
On separate note, the "Changelist" part in the PR Template is not edited accordingly. It is a common best practice to include a short description of the PR in the first comment.

image

PR #204, opened 2025-04-17
LONDON_JAN25 | KHALIL ALI | MODULE_DATA_FLOWS | SPRINT 2 | BOOK_LIBRARY

khalil-ali1:
## Learners, PR Template

Self checklist

## Changelist

Fix all book library project bugs.

## Questions

### DIFF
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -6,7 +6,7 @@ window.addEventListener("load", function (e) {
 });
 
 function populateStorage() {
-  if (myLibrary.length == 0) {
+  if (myLibrary.length === 0) {
     let book1 = new Book("Robison Crusoe", "Daniel Defoe", "252", true);
     let book2 = new Book(
       "The Old Man and the Sea",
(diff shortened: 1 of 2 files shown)


debugging/book-library/script.js
```diff
@@ -54,7 +54,7 @@ function render() {
   let table = document.getElementById("display");
   let rowsNumber = table.rows.length;
   //delete old table
-  for (let n = rowsNumber - 1; n > 0; n-- {
+  for (let n = rowsNumber - 1; n > 0; n--) {
     table.deleteRow(n);
   }
```
cjyuan, line 59: Instead of deleting the table rows one by one, can you think of a more efficient way to remove all rows (except the `<th>...</th>`) in the table?

debugging/book-library/script.js
cjyuan, line 1: - In terms of input validation, 
  1. Do you want to allow book title and author name to contain only space characters?

- Lines 75, 93:

- Line 69:
  - The current method of assigning book titles to HTML elements can cause display issues if a title contains special character sequences like `<i>`.

cjyuan (commented):

I also suggested some improvements for you to consider.

khalil-ali1:
I used [ https://validator.w3.org/](url) to check my HTML file.
I fixed the code and got the "No errors or warnings to show" message.

debugging/book-library/script.js
khalil-ali1: - i. I have checked all input properly
- ii. `.value` can not be null.
- iii.` .trim()` deletes all spaces
- iv. the "pages" input should not be empty and should be  a positive integer


id attributes are not unique
There is no need to assign id attributes  to either buttons

you are right. I used `textContent ` instead of `innerhtml`

PR #208, opened 2025-04-20
WESTMIDLANDS-ITP-JAN|SEGUN FOLAYAN|DATA FLOWS|BOOK-LIBRARY 

segunfolayan:
## Learners, PR Template

Self checklist

- [X] I have titled my PR with COHORT_NAME | FIRST_NAME LAST_NAME | REPO_NAME | WEEK 
- [X] My changes follow the [style guide](https://curriculum.codeyourfuture.io/guides/contributing/)
- [ X] My changes meet the [requirements](./README.md) of this task

## Changelist

BOOK LIBRARY PROJECT PR

## Questions

### DIFF
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -31,8 +31,12 @@ function submit() {
   if (
     title.value == null ||
     title.value == "" ||
+    author.value == null ||
+    author.value == "" ||
     pages.value == null ||
-    pages.value == ""
+    pages.value == ""||
+    check.checked == null||
+    check.checked == ""
   ) {
     alert("Please fill all fields!");
     return false;
@@ -40,6 +44,10 @@ function submit() {
     let book = new Book(title.value, author.value, pages.value, check.checked);
     myLibrary.push(book);
     render();
+    title.value = "";
+    author.value = "";
+    pages.value = "";
+    check.checked = false;
   }
 }
 
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -16,7 +16,7 @@ function populateStorage() {
     );
     myLibrary.push(book1);
     myLibrary.push(book2);
-    render;
+    //render;
   }
 }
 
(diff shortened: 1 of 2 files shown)


debugging/book-library/script.js
```diff
@@ -76,7 +84,7 @@ function render() {
     changeBut.className = "btn btn-success";
     wasReadCell.appendChild(changeBut);
     let readStatus = "";
-    if (myLibrary[i].check == false) {
+    if (myLibrary[i].check == true) {
```
jenny-alexander, line 87: Can you think of a way to rewrite lines 87-91 in order to reduce repetition?
- The value within myLibrary[i].check will hold either true or false. 
- Can you use a ternary operator to achieve the same result in one line of code 

_The syntax of the ternary operator is like this: 
condition ? trueExpression : falseExpression. 
If the condition evaluates to true, the expression after the ? is executed; otherwise, the expression after the : is executed._

jenny-alexander (changes requested):
Your PR looks good 👍

Here are a few change requests that will make it even better:
1. Can you review all of the places that you declared a variable with 'let'? In general, you will mostly always use const (and this should be your default). If you need to reassign a value to the variable, then you would use let. Here is a video you can watch: https://youtu.be/RE6qf3As-XU?si=_YaiEPeBC5Cx_j56

2. The 4 variables on lines 24-27 are currently declared as global variables. However, they are only used within the submit function. 
Can modify your code so that they are no longer global variables and are declared as local variables?
Making this change means the variables won't be created until they are needed within the submit function. This makes the code cleaner and more maintainable!
https://www.w3schools.com/js/js_scope.asp

debugging/book-library/script.js
segunfolayan: readStatus = myLibrary[i].check ? "Yes" : "No";
  I have replaced with this line of code.
  Thank you

segunfolayan:
Hello,
I will like to let you know that I have implemented the changes as advised.
Thank you.

jenny-alexander (approved):

PR #222, opened 2025-04-23
ITP-Jan| NW | Jovy So | Module-Data-Flows | Book-library

jovyso:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -16,8 +16,8 @@ function populateStorage() {
     );
     myLibrary.push(book1);
     myLibrary.push(book2);
-    render();
   }
+  render(); 
 }
 
 const title = document.getElementById("title");
@@ -25,8 +25,6 @@ const author = document.getElementById("author");
 const pages = document.getElementById("pages");
 const check = document.getElementById("check");
 
-//check the right input from forms and if its ok -> add the new book (object in array)
-//via Book function and start render function
 function submit() {
   if (
     title.value == null ||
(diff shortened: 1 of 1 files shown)


debugging/book-library/script.js
```diff
@@ -16,17 +16,15 @@ function populateStorage() {
     );
     myLibrary.push(book1);
     myLibrary.push(book2);
-    render();
   }
+  render();
```
cjyuan, line 20: If we call `render()` here, do we still need calling `render()` at line 5?

debugging/book-library/script.js
```diff
@@ -53,11 +51,9 @@ function Book(title, author, pages, check) {
 function render() {
   let table = document.getElementById("display");
   let rowsNumber = table.rows.length;
-  //delete old table
-  for (let n = rowsNumber - 1; n > 0; n-- {
+  for (let n = rowsNumber - 1; n > 0; n--) { 
     table.deleteRow(n);
   }
```
cjyuan, line 56: Instead of deleting the table rows one by one, can you think of a more efficient way to remove all rows (except the `<th>...</th>`) in the table?

debugging/book-library/script.js
cjyuan, line 1: In terms of input validation (lines 30-36), 
1. Do you want to allow book title and author name to contain only space characters? For example, `"    "`.


- Lines 70, 87:

- Line 65:
  - The current method of assigning book titles to HTML elements can cause display issues if a title contains special character sequences like `<i>`.

- Can you suggest a more consistent naming convention for the variables representing the two buttons, currently named `changeBut` and `delButton`?

- Line 74-79: Can you use the conditional operator, `? :` to simplify these statements into one line of code?

cjyuan (commented):

The code in `script.js` can also use some improvement.

PR #238, opened 2025-07-25
London | ITP-May-25 | Jamal Laqdiem | Module-Data-Flows | Sprint1-Feature/destructuring

jamallaqdiem:
## Learners, PR Template

Self checklist

- [ ] I have titled my PR with Region | Cohort | FirstName LastName | Sprint | Assignment Title
- [ ] My changes follow the [style guide](https://curriculum.codeyourfuture.io/guides/reviewing/style-guide/)

## Changelist
 use object destructuring to extract values from an object and assign them to variables 
Use object destructuring to extract the values you need out of each element in the array
Use object destructuring to access the values you need from each item.
## Questions

### DIFF
--- a/Sprint-1/destructuring/exercise-3/exercise.js
+++ b/Sprint-1/destructuring/exercise-3/exercise.js
@@ -6,3 +6,34 @@ let order = [
   { itemName: "Hot Coffee", quantity: 2, unitPricePence: 100 },
   { itemName: "Hash Brown", quantity: 4, unitPricePence: 40 },
 ];
+
+
+const quantityWidth = 9
+const itemWidth = 18
+
+const headQuantity = 'QTY'.padEnd(9)
+const headItem = 'ITEM'.padEnd(18)
+
+console.log(`${headQuantity}${headItem} TOTAL`)
+
+const isOrder =  (number) => typeof number.quantity==='number'
+order.forEach((item) => {
+
+const totalUnit = item.quantity*item.unitPricePence
+const orderNumber = Number(item.quantity)
+const nameItem = String(item.itemName)
+
+const orderNumberAlign = String(orderNumber).padEnd(9)
+const nameItemAlign= nameItem.padEnd(18)
+
+console.log(`${orderNumberAlign} ${nameItemAlign}${totalUnit}`)
+});
+
+
+
+const total = order.reduce((accumulator,item) =>{
+  itemTotal =item.quantity*item.unitPricePence
+  return accumulator + itemTotal
+},0)
+
+console.log(`Total: ${total}`)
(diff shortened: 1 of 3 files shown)


Sprint-1/destructuring/exercise-3/exercise.js
```diff
@@ -6,3 +6,34 @@ let order = [
   { itemName: "Hot Coffee", quantity: 2, unitPricePence: 100 },
   { itemName: "Hash Brown", quantity: 4, unitPricePence: 40 },
 ];
+
+
+const quantityWidth = 9
```
LonMcGregor, line 11: you defined the formatting variables here which is a good idea, but then you never use them. can you fix that?

Sprint-1/destructuring/exercise-3/exercise.js
```diff
@@ -6,3 +6,34 @@ let order = [
...
+
+const headQuantity = 'QTY'.padEnd(9)
+const headItem = 'ITEM'.padEnd(18)
+
+console.log(`${headQuantity}${headItem} TOTAL`)
+
+const isOrder =  (number) => typeof number.quantity==='number'
+order.forEach((item) => {
+
+const totalUnit = item.quantity*item.unitPricePence
```
LonMcGregor, line 22: this is a block of code - for readability you should indent it

Sprint-1/destructuring/exercise-3/exercise.js
```diff
@@ -6,3 +6,34 @@ let order = [
...
+const quantityWidth = 9
+const itemWidth = 18
+
+const headQuantity = 'QTY'.padEnd(9)
+const headItem = 'ITEM'.padEnd(18)
+
+console.log(`${headQuantity}${headItem} TOTAL`)
+
+const isOrder =  (number) => typeof number.quantity==='number'
+order.forEach((item) => {
```
LonMcGregor, line 20: Can you write this function using object destructuring? The readme says this is a requirement of the task.

LonMcGregor (commented):
Good start on this task - Remember what I have said previously in other tasks to be consistent with semicolons when the project already uses them. There are a few issues in exercise 3 here for you to look at

Sprint-1/destructuring/exercise-3/exercise.js
jemaljemy: Thanks for reviewing, I noticed that i forgot to delete the two variables they are redundant as I had assigned the value in the following lines.

Sprint-1/destructuring/exercise-3/exercise.js
jemaljemy: Done.

LonMcGregor:
Great work - you're done with this task

PR #241, opened 2025-07-25
London | May 2025 | Sisay Mehari | Sprint 2 | Book Library Debugging 

Sisu860:
Learners, PR Template

Self checklist

### ✅ What I’ve Done

This pull request addresses multiple bugs in the Book Library project:

#### 🔧 Bug Fixes

1. **Books not showing on load**
   - Called `render()` after `populateStorage()` to ensure books display correctly when the page loads.

2. **Error in console when adding a book**
   - Replaced incorrect `library.push(...)` with the correct `myLibrary.push(...)` in the `submit()` function.

3. **Title used instead of author**
   - Corrected the `Book()` constructor arguments to pass `author.value` instead of mistakenly using `title.value` twice.

No question at the moment.

### DIFF
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -37,8 +37,8 @@ function submit() {
     alert("Please fill all fields!");
     return false;
   } else {
-    let book = new Book(title.value, title.value, pages.value, check.checked);
-    library.push(book);
+    let book = new Book(title.value, author.value, pages.value, check.checked);
+    myLibrary.push(book);
     render();
   }
 }
@@ -54,7 +54,7 @@ function render() {
   let table = document.getElementById("display");
   let rowsNumber = table.rows.length;
   //delete old table
-  for (let n = rowsNumber - 1; n > 0; n-- {
+  for (let n = rowsNumber - 1; n > 0; n-- ) {
     table.deleteRow(n);
   }
   //insert updated row and cells
@@ -76,7 +76,7 @@ function render() {
     changeBut.className = "btn btn-success";
     wasReadCell.appendChild(changeBut);
     let readStatus = "";
-    if (myLibrary[i].check == false) {
+    if (myLibrary[i].check == true) {
       readStatus = "Yes";
     } else {
       readStatus = "No";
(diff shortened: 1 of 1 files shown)


debugging/book-library/script.js
```diff
@@ -76,7 +76,7 @@ function render() {
     changeBut.className = "btn btn-success";
     wasReadCell.appendChild(changeBut);
     let readStatus = "";
-    if (myLibrary[i].check == false) {
+    if (myLibrary[i].check == true) {
```
LonMcGregor, line 79: if `.check` is a boolean, do you need to compare it to `== true`?

debugging/book-library/script.js
```diff
@@ -37,8 +37,8 @@ function submit() {
     alert("Please fill all fields!");
```
LonMcGregor, line 37: Is there any input validation missing from the pages? What about the author?

debugging/book-library/script.js
```diff
@@ -90,11 +90,11 @@ function render() {
...
-    deleteCell.appendChild(delBut);
-    delBut.className = "btn btn-warning";
-    delBut.innerHTML = "Delete";
-    delBut.addEventListener("clicks", function () {
+    delButton.id = i + 5;
+    deleteCell.appendChild(delButton);
+    delButton.className = "btn btn-warning";
+    delButton.innerHTML = "Delete";
+    delButton.addEventListener("click", function () {
       alert(`You've deleted title: ${myLibrary[i].title}`);
```
LonMcGregor, line 98: can you think of a better user interaction than `alert()` here? Deleting something can be pretty bad, and being notified after a delete has happened isnt usually what you see.

LonMcGregor (commented):
Good work on debugging the problems. Remember the readme says there are other bugs than the ones mentioned. I've left some pointers for where you might find them

debugging/book-library/script.js
Sisu860: I have done all the input validation.

debugging/book-library/script.js
Sisu860: We do not need to compare it.

debugging/book-library/script.js
Sisu860: I have added a warning to notify before deleting it.

LonMcGregor:
Great work on this -you are done with this task now

PR #252, opened 2025-07-29
London| ITP-May-2025| Zohreh Kazemianpour |Book library

zohrehKazemianpour:
## Learners, PR Template

Self checklist

- [ ] I have titled my PR with Region | Cohort | FirstName LastName | Sprint | Assignment Title
- [ ] My changes follow the [style guide](https://curriculum.codeyourfuture.io/guides/reviewing/style-guide/)

## Changelist

This PR addresses all major functionality bugs in the Book Library:

✔ Fixed book display
Books now render correctly on page load
The default books appear as intended

✔ Corrected form submission
Properly captures all fields (title, author, pages, read status)
Fixed title/author mixup bug
Added validation for required fields

✔ Fixed delete functionality
Delete buttons now work reliably
Uses stable data-index system instead of fragile IDs

✔ Read status accuracy
Correctly saves and displays read/unread state
Toggle buttons update properly

## Questions

### DIFF
--- a/debugging/book-library/index.html
+++ b/debugging/book-library/index.html
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <title> </title>
+    <title>Book Library </title>
     <meta
       charset="utf-8"
       name="viewport"
--- a/debugging/book-library/index.html
+++ b/debugging/book-library/index.html
@@ -64,8 +64,8 @@
         <input
           type="submit"
           value="Submit"
+          id="submit"
           class="btn btn-primary"
-          onclick="submit();"
         />
       </div>
     </div>
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -7,7 +7,7 @@ window.addEventListener("load", function (e) {
 
 function populateStorage() {
   if (myLibrary.length === 0) {
-    let book1 = new Book("Robison Crusoe", "Daniel Defoe", "252", true);
+    let book1 = new Book("Robinson Crusoe", "Daniel Defoe", "252", true);
     let book2 = new Book("The Old Man and the Sea","Ernest Hemingway","127", true);
     myLibrary.push(book1);
     myLibrary.push(book2);
@@ -76,7 +76,7 @@ function render() {
 
     //add and wait for action for read/unread button
     let changeBut = document.createElement("button");
-    changeBut.id = i;
+    changeBut.dataset.index = i;
     changeBut.className = "btn btn-success";
     wasReadCell.appendChild(changeBut);
     let readStatus = "";
@@ -88,19 +88,21 @@ function render() {
     changeBut.innerText = readStatus;
 
     changeBut.addEventListener("click", function () {
-      myLibrary[i].check = !myLibrary[i].check;
+      const index = parseInt(this.dataset.index);
+      myLibrary[index].check = !myLibrary[index].check;
       render();
     });
 
     //add delete button to every row and render again
     let delBut = document.createElement("button");
-    delBut.id = i + 5;
+    delBut.dataset.index = i; 
     deleteCell.appendChild(delBut);
     delBut.className = "btn btn-warning";
     delBut.innerHTML = "Delete";
     delBut.addEventListener("click", function () {
       alert(`You've deleted title: ${myLibrary[i].title}`);
-      myLibrary.splice(i, 1);
+      const index = parseInt(this.dataset.index)
+      myLibrary.splice(index, 1);
       render();
     });
   }
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -23,19 +23,24 @@ const check = document.getElementById("check");
 //via Book function and start render function
 function submit() {
   if (
-    title.value == null ||
-    title.value == "" ||
-    pages.value == null ||
-    pages.value == ""
+    title.value.trim() === "" ||
+    author.value.trim() === "" ||
+    pages.value === ""
   ) {
     alert("Please fill all fields!");
     return false;
   } else {
     let book = new Book(title.value, author.value, pages.value, check.checked);
-    library.push(book);
+    myLibrary.push(book);
   
   }
 }
+const submitBtn = document.getElementById("submit");
+submitBtn.addEventListener("click", function(e) {
+  e.preventDefault(); 
+  submit();
+  render();
+});
 
 function Book(title, author, pages, check) {
   this.title = title;
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -40,6 +40,10 @@ submitBtn.addEventListener("click", function(e) {
   e.preventDefault(); 
   submit();
   render();
+  title.value = "";
+  author.value = "";
+  pages.value = "";
+  check.checked = false;
 });
 
 function Book(title, author, pages, check) {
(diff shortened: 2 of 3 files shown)


debugging/book-library/index.html

debugging/book-library/script.js
```diff
@@ -6,18 +6,12 @@ window.addEventListener("load", function (e) {
...
-    let book1 = new Book("Robison Crusoe", "Daniel Defoe", "252", true);
-    let book2 = new Book(
-      "The Old Man and the Sea",
-      "Ernest Hemingway",
-      "127",
-      true
-    );
+  if (myLibrary.length === 0) {
+    let book1 = new Book("Robinson Crusoe", "Daniel Defoe", "252", true);
+    let book2 = new Book("The Old Man and the Sea","Ernest Hemingway","127", true);
```
cjyuan, line 11: Did you notice it was weird to pass the value of `page` as a string instead of as a number in the original code?

debugging/book-library/script.js
```diff
@@ -29,19 +23,28 @@ const check = document.getElementById("check");
...
+    author.value.trim() === "" ||
+    pages.value === ""
   ) {
     alert("Please fill all fields!");
     return false;
   } else {
-    let book = new Book(title.value, title.value, pages.value, check.checked);
-    library.push(book);
-    render();
+    let book = new Book(title.value, author.value, pages.value, check.checked);
```
cjyuan, line 33: - Do you want to keep leading and trailing space characters in title, author, and pages?

- The following page numbers are possible
image

debugging/book-library/script.js
```diff
@@ -54,13 +57,13 @@ function render() {
   let table = document.getElementById("display");
   let rowsNumber = table.rows.length;
   //delete old table
-  for (let n = rowsNumber - 1; n > 0; n-- {
+  for (let n = rowsNumber - 1; n > 0; n--) {
     table.deleteRow(n);
   }
```
cjyuan, line 62: Instead of deleting the table rows one by one, can you think of a more efficient way to remove all rows (except the `<th>...</th>`) in the table?

debugging/book-library/script.js
```diff
@@ -54,13 +57,13 @@ function render() {
...
   //insert updated row and cells
   let length = myLibrary.length;
   for (let i = 0; i < length; i++) {
-    let row = table.insertRow(1);
+    let row = table.insertRow(i+1);
```
cjyuan, line 66: We could also omit the argument in `insertRow()` if we wanted to append a row to the table.

debugging/book-library/script.js
```diff
@@ -69,34 +72,32 @@ function render() {
     titleCell.innerHTML = myLibrary[i].title;
     authorCell.innerHTML = myLibrary[i].author;
     pagesCell.innerHTML = myLibrary[i].pages;
```
cjyuan, line 74: Please note that when setting the text content of an HTML element, there are subtle but important differences between using `.innerHTML`, `innerText`, and `textContent`.

The current method of assigning book titles to HTML elements can cause display issues if a title contains special character sequences like `<i>`.

debugging/book-library/script.js
```diff
@@ -69,34 +72,32 @@ function render() {
...
     changeBut.className = "btn btn-success";
     wasReadCell.appendChild(changeBut);
-    let readStatus = "";
-    if (myLibrary[i].check == false) {
-      readStatus = "Yes";
-    } else {
-      readStatus = "No";
-    }
+    readStatus = myLibrary[i].check ? "Yes" : "No";
     changeBut.innerText = readStatus;
```
cjyuan, line 83: The declaration of variable `readStatus` was deleted. You could also combined lines 82-83 into one statement without using `readStatus`.

debugging/book-library/script.js
```diff
@@ -69,34 +72,32 @@ function render() {
...
+    delBut.dataset.index = i; 
     deleteCell.appendChild(delBut);
     delBut.className = "btn btn-warning";
     delBut.innerHTML = "Delete";
-    delBut.addEventListener("clicks", function () {
+    delBut.addEventListener("click", function () {
       alert(`You've deleted title: ${myLibrary[i].title}`);
-      myLibrary.splice(i, 1);
+      const index = parseInt(this.dataset.index)
+      myLibrary.splice(index, 1);
```
cjyuan, line 100: The alert message is shown before the book is actually deleted; the deletion only occurs after the alert dialog is dismissed. This introduces a risk that the operation may not complete (e.g., if the user closes the browser before dismissing the alert).

debugging/book-library/index.html
zohrehKazemianpour: Thank you for the W3C HTML Validator link. It's a very useful tool! I've fixed all the validation errors in the index.html file. The main issues were invalid input types, missing lang attribute, and improper meta tag structure. The HTML now passes validation completely.

debugging/book-library/script.js
zohrehKazemianpour: Thank you for pointing this out! Initially, I didn't notice anything wrong with storing page value as a string, but after your comment, I did some research and I learned that it is best practice to store variables based on their correct data types to prevent bugs and enable operations later on. However, I still found it confusing because based on my understanding, even with string app was functional and working. So it led me to learn more about data flow. I learned that HTML inputs always return strings (even for number inputs). We can choose to convert and store data in the appropriate type. The display layer automatically converts numbers back to strings for rendering, and this creates a clear separation between data storage (numbers) and presentation (strings).

zohrehKazemianpour:
Hi @cjyuan, thank you so much as always, for the valuable feedback. I have updated the code based on your comments. Could you please review when you have a moment? I appreciate your insights!

debugging/book-library/script.js
```diff
@@ -29,15 +29,18 @@ function submit() {
...
     alert("Please fill all fields!");
     return false;
-  } else {
-    let book = new Book(title.value, author.value, pages.value, check.checked);
+  } 
+  if (isNaN(parseInt(pages.value)) || parseInt(pages.value) <= 0) {
+    alert("Please enter a valid positive number for pages!");
+    return false;
+  }  
+    let book = new Book(title.value.trim(), author.value.trim(), parseInt(pages.value), check.checked);
```
cjyuan, line 37: For better performance (reduce number of function calls) and reducing the chance of using raw input accidently, we could stored the pre-processed/sanitized/normalized input in some variables first, and reference the variables in other part of the function.
```javascript
  const cleanTitle = title.value.trim();
  ...
  
  const pageNum = parseInt(pages.value);
  ...
  
  const book = new Book(cleanTitle, cleanAuthor, pageNum, check.checked);

```

debugging/book-library/script.js
```diff
@@ -7,11 +7,11 @@ window.addEventListener("load", function (e) {
...
-    let book2 = new Book("The Old Man and the Sea","Ernest Hemingway","127", true);
+    let book1 = new Book("Robinson Crusoe", "Daniel Defoe", 252, true);
+    let book2 = new Book("The Old Man and the Sea", "Ernest Hemingway", 127, true);
     myLibrary.push(book1);
     myLibrary.push(book2);
- }
+  }
 }
 
 const title = document.getElementById("title");
```
cjyuan, line 17: Using descriptive and consistent suffixes (like `El`, `Input`, `Btn`, `Form`, etc.) for variables that store DOM elements can improve code readability and maintainability.

cjyuan (commented):
Changes look great! Well done!

zohrehKazemianpour:
> Changes look great! Well done!

Thank you so much for the complete label. I appreciate the time and energy you put into reviewing our coursework.

PR #253, opened 2025-07-29
London | ITP-May-25 | Emiliano Uruena | Data-Flows Sprint-2 | Book-Library 📚 

Emilianouz:
Self checklist

## Changelist

**Bug Fixes Implemented in JavaScript and HTML Functionality:**

- [x] View a list of books I've read:

The display function now correctly renders the full list of books stored in memory or local storage. Any updates to the list—such as adding or removing books—are reflected immediately in the user interface.

- [x] Add new books to the list:

Users can now successfully add books they've read to the list. Each book entry includes:

- Title
- Author
- Number of pages
- A checkbox to indicate whether the book has been read

All form inputs are now properly validated. If any required field is missing or left empty, the book will not be added, and the system will display an alert message prompting the user to complete all fields. This prevents empty or incomplete entries in the book list.

- [x] Remove books from the list:

Each book in the list now includes a "Remove" button. When clicked, the corresponding book is deleted from the list both visually and from the underlying data. The removal process has been debugged to ensure that it consistently targets the correct entry without affecting others.

### DIFF
--- a/debugging/book-library/index.html
+++ b/debugging/book-library/index.html
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <title> </title>
+    <title>Book Library</title>
     <meta
       charset="utf-8"
       name="viewport"
@@ -73,10 +73,10 @@
     <table class="table" id="display">
       <thead class="thead-dark">
         <tr>
-          <th id="title">Title</th>
-          <th id="author">Author</th>
-          <th id="pages">Number of Pages</th>
-          <th id="check">Read</th>
+          <th id="tab-title">Title</th>
+          <th id="tab-author">Author</th>
+          <th id="tab-pages">Number of Pages</th>
+          <th id="tab-check">Read</th>
           <th></th>
         </tr>
       </thead>
--- a/debugging/book-library/index.html
+++ b/debugging/book-library/index.html
@@ -1,12 +1,9 @@
 <!DOCTYPE html>
-<html>
+<html lang="en">
   <head>
     <title>Book Library</title>
-    <meta
-      charset="utf-8"
-      name="viewport"
-      content="width=device-width, initial-scale=1.0"
-    />
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />  
     <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.4.1/jquery.min.js"></script>
     <script src="https://cdnjs.cloudflare.com/ajax/libs/popper.js/1.16.0/umd/popper.min.js"></script>
     <script src="https://maxcdn.bootstrapcdn.com/bootstrap/4.4.1/js/bootstrap.min.js"></script>
@@ -31,7 +28,7 @@
       <div class="form-group">
         <label for="title">Title:</label>
         <input
-          type="title"
+          type="text"
           class="form-control"
           id="title"
           name="title"
@@ -39,7 +36,7 @@
         />
         <label for="author">Author: </label>
         <input
-          type="author"
+          type="text"
           class="form-control"
           id="author"
           name="author"
@@ -81,13 +78,6 @@
         </tr>
       </thead>
       <tbody>
-        <tr>
-          <td></td>
-          <td></td>
-          <td></td>
-          <td></td>
-          <td></td>
-        </tr>
       </tbody>
     </table>
 
--- a/debugging/book-library/index.html
+++ b/debugging/book-library/index.html
@@ -73,10 +73,10 @@
     <table class="table" id="display">
       <thead class="thead-dark">
         <tr>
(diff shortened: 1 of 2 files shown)


debugging/book-library/index.html

debugging/book-library/script.js
```diff
@@ -31,14 +31,16 @@ function submit() {
   if (
     title.value == null ||
     title.value == "" ||
+    author.value == null ||
+    author.value == "" ||
     pages.value == null ||
     pages.value == ""
```
cjyuan, line 37: In terms of input validation, 
1. Do you want to allow book title and author name to contain only space characters?

debugging/book-library/script.js
cjyuan, line 1: - On lines 10-14: Did you notice it was weird to pass the value of `page` as a string instead of as a number in the original code?

- On lines 23-26:   Using descriptive and consistent suffixes (like `El`, `Input`, `Btn`, `Form`, etc.) for variables that store DOM elements can improve code readability and maintainability.

- Currently on page load, `render()` is unnecessarily executed twice. Can you address this issue?

debugging/book-library/script.js
```diff
@@ -31,14 +31,16 @@ function submit() {
...
   ) {
     alert("Please fill all fields!");
     return false;
   } else {
-    let book = new Book(title.value, title.value, pages.value, check.checked);
-    library.push(book);
+    let book = new Book(title.value, author.value, pages.value, check.checked);
```
cjyuan, line 42: Using user input directly without sanitization, normalization, or conversion to the proper data type can be dangerous.

Can you ensure the arguments passed to the Book constructor are properly **pre-processed**?

debugging/book-library/script.js
```diff
@@ -54,7 +56,7 @@ function render() {
   let table = document.getElementById("display");
   let rowsNumber = table.rows.length;
   //delete old table
-  for (let n = rowsNumber - 1; n > 0; n-- {
+  for (let n = rowsNumber - 1; n > 0; n--) {
     table.deleteRow(n);
   }
```
cjyuan, line 61: To clear the table before repopulating it, instead of deleting the table rows one by one, can you think of a more efficient way to remove all rows (except the `<th>...</th>`) in the table?

debugging/book-library/script.js
```diff
@@ -76,7 +78,7 @@ function render() {
     changeBut.className = "btn btn-success";
     wasReadCell.appendChild(changeBut);
     let readStatus = "";
-    if (myLibrary[i].check == false) {
+    if (myLibrary[i].check === true) {
       readStatus = "Yes";
     } else {
       readStatus = "No";
```
cjyuan, line 84: To practice using the `? :` conditional operator, can you refactor the code on lines 80-86 into a single line of code ?

debugging/book-library/script.js
```diff
@@ -89,12 +91,12 @@ function render() {
     });
 
     //add delete button to every row and render again
-    let delButton = document.createElement("button");
+    let delBut = document.createElement("button");
     delBut.id = i + 5;
     deleteCell.appendChild(delBut);
     delBut.className = "btn btn-warning";
     delBut.innerHTML = "Delete";
```
cjyuan, line 98: Please note that when setting the text content of an HTML element, there are subtle but important differences between using `.innerHTML`, `innerText`, and `textContent`.

- The current method of assigning book titles to HTML elements (on lines 71) can cause display issues if a title contains special character sequences like `<i>`.

- Did you notice that on line 86, the string is assigned to `.innerText` but on line 98, the string value is assigned to `.innerHTML`? Which property should we use?

debugging/book-library/script.js
```diff
@@ -89,12 +91,12 @@ function render() {
...
-    delBut.addEventListener("clicks", function () {
+    delBut.addEventListener("click", function () {
       alert(`You've deleted title: ${myLibrary[i].title}`);
       myLibrary.splice(i, 1);
```
cjyuan, line 101: The alert message is shown before the book is actually deleted; the deletion only occurs after the alert dialog is dismissed. This introduces a risk that the operation may not complete (e.g., if the user closes the browser before dismissing the alert).

debugging/book-library/script.js
Emilianouz: I though that the number of pages no necessarily had to be number because there are not math operations to do. But, yes it was weird.

Emilianouz (commented):
Issues solved according to the feedback.
- Using user input with validation.
- Arguments are pre-processed/sanitize before calling the constructor; removing leading/trailing whitespace
- rows are inserted inside the <tbody>, not the <thead>

debugging/book-library/index.html
Emilianouz: Apologies, I’ve reviewed and resolved it.

Emilianouz (commented):
It were reviewed and resolved.

PR #278, opened 2025-08-06
Glasgow | ITP-May-2025 | Nataliia Volkova | Sprint 2 | book library

Nataliia74:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -27,29 +27,25 @@ const check = document.getElementById("check");
 //check the right input from forms and if its ok -> add the new book (object in array)
 //via Book function and start render function
 function bookSubmit() {
-  const pagesFloat = parseFloat(pagesInput.value);
+  let bookPagesFloat = parseFloat(pagesInput.value);
+  let bookTitle = titleInput.value.toString().trim();
+  let bookAuthor = authorInput.value.toString().trim();
+  let bookCheck = check.checked;
 
   if (
-    titleInput.value.toString().trim() === "" ||
-    authorInput.value.toString().trim() === "" ||
-    !Number.isInteger(pagesFloat) ||
-    pagesFloat <= 0
+    bookTitle === "" ||
+    bookAuthor === "" ||
+    !Number.isInteger(bookPagesFloat) ||
+    bookPagesFloat <= 0
   ) {
     alert("Please fill all fields!");
     return false;
   } else {
-    let book = new Book(
-      titleInput.value,
-      authorInput.value,
-      pagesFloat,
-      check.checked
-    );
+    let book = new Book(bookTitle, bookAuthor, bookPagesFloat, bookCheck);
     myLibrary.push(book);
     render();
-    (titleInput.value = ""),
-      (authorInput.value = ""),
-      (pagesInput.value = ""),
-      (check.checked = false);
+    titleInput.value = authorInput.value = pagesInput.value = "";
+    check.checked = false;
   }
 }
 
@@ -61,8 +57,6 @@ function Book(title, author, pages, check) {
 }
 
 function render() {
-  console.count("render called");
-
   let table = document.getElementById("display");
   table.tBodies[0].innerHTML = "";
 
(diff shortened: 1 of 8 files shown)


debugging/book-library/script.js
```diff
@@ -7,42 +7,48 @@ window.addEventListener("load", function (e) {
...
+    !Number.isInteger(pagesFloat) ||
+    pagesFloat <= 0
   ) {
     alert("Please fill all fields!");
     return false;
   } else {
-    let book = new Book(title.value, author.value, pages.value, check.checked);
+    let book = new Book(
+      titleInput.value,
+      authorInput.value,
```
cjyuan, line 43: These raw input values could include leading and trailing space characters.

To reduce the chance of using raw input accidently, we could stored the pre-processed/sanitized/normalized input in some variables first, and reference the variables in other part of the function.

debugging/book-library/script.js
```diff
@@ -7,42 +7,48 @@ window.addEventListener("load", function (e) {
...
+    );
     myLibrary.push(book);
     render();
-    (title.value = ""),
-      (author.value = ""),
-      (pages.value = ""),
+    (titleInput.value = ""),
+      (authorInput.value = ""),
+      (pagesInput.value = ""),
       (check.checked = false);
```
cjyuan, line 52: Why not express them in one of the following forms?
```
  A = "";
  B = "";
  C = "";
```
or
```
  A = B = C = "";
```

debugging/book-library/script.js
```diff
@@ -55,53 +61,65 @@ function Book(title, author, pages, check) {
 }
 
 function render() {
+  console.count("render called");
```
cjyuan, line 64: PR submissions should be free of unnecessary code, such as debugging code.

debugging/book-library/script.js
```diff
@@ -55,53 +61,65 @@ function Book(title, author, pages, check) {
...
     } else {
       readStatus = "No";
     }
-    changeBut.innerText = readStatus;
-
-    changeBut.addEventListener("click", function () {
-      myLibrary[i].check = !myLibrary[i].check;
-      render();
-    });
+    changeButton.innerText = readStatus;
```
cjyuan, line 92: This could be a good opportunity to practice using the `? :` conditional operator. Can you rewrite the code on lines 86–92 as a single statement?

debugging/book-library/script.js
```diff
@@ -55,53 +61,65 @@ function Book(title, author, pages, check) {
...
+  if (action === "toggle-book") {
+    myLibrary[index].check = !myLibrary[index].check;
+    render();
+  }
+  if (action === "delete-book") {
+    const deletion = myLibrary[index].title;
+    myLibrary.splice(index, 1);
+    render();
+    setTimeout(() => {
+      confirm(`${deletion} has been successfully deleted!`);
```
cjyuan, line 122: A "confirm" dialog box has "Yes" and "No" buttons. How should a user interpret these two buttons after the book has been deleted?

cjyuan (commented):
Changes look good. I have a few more suggestions.

debugging/book-library/script.js
Nataliia74: Done.

debugging/book-library/script.js
Nataliia74: I switched to the "alert" dialogue box. Left the "confirm" one when I was experimenting with code

debugging/book-library/script.js
Nataliia74: That's taught me new things about chaining variables. I can chain variables only if we expect they will be assigned to the same type of values, and they are not local variables that reference appropriate DOM elements. That means chaining local variables is not safe because I can accidentally overwrite something important. It is safer to chain properties, like `.value, .textContent, .checked`.

debugging/book-library/script.js
Nataliia74: Thank you for teaching me how to write readable code.

debugging/book-library/script.js
cjyuan: `A = B = C = ""` is evaluated as `A = (B = (C = ""))` (The assignment operator is evaluated from right to left).

The expression `C = ""` assigns `""` to `C` and evaluates to the value of `C`.
The value of the expression `C = ""` is then assigned to B, and then the value of the expression `B = C = ""` is assigned to `A`.

It is just a handy syntax to assign a value to multiple variables. 

I don't quite follow what you meant by
>... they are not local variables that reference appropriate DOM elements". That means chaining local variables is not safe because I can accidentally overwrite something important. It is safer to chain properties, like .value, .textContent, .checked.

Primitive type values (including String) in JS are immutable so it is safe to assign the same primitive value to multiple variables.

Objects (including arrays) are shared by reference. So
```
A = [];    // A and B are assigned different arrays
B = [];
```
is not the same as
```
A = B = [];   // A and B share the same array
```

As long as we recognize exactly what the code does, we don't have to avoid using the 2nd syntax.

cjyuan:
Changes look good. Well done.

PR #292, opened 2025-08-10
LONDON | MAY2025 |FATMA_DEGIRMENCI | SPRINT2 | BOOK_LIBRARY

fatmaevin:
Self checklist

### DIFF
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -3,11 +3,9 @@ let myLibrary = [];
 window.addEventListener("load", function (e) {
   loadLibrary();
   if(myLibrary.length===0){
-  populateStorage();
-  
-}
+    populateStorage();
+  }
   render();
-
 });
 
 function populateStorage() {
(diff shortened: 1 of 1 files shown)


fatmaevin:
I tried to improve and adjust my code according to the feedback. Thank you.

debugging/book-library/script.js
```diff
@@ -1,45 +1,72 @@
...
+}
+
+function protectMyLibrary(){
+  localStorage.setItem("myLibrary" , JSON.stringify(myLibrary));
+}
+
+function loadLibrary() {
+  const storedLibrary = localStorage.getItem("myLibrary");
+  if (storedLibrary) {
+    myLibrary = JSON.parse(storedLibrary);
```
cjyuan, line 69: When a book is added to the array in `submit()` and `populateStorage()`, the code adds an instance of Book to the array.
On line 63, only the key–value pairs of the object are stored in the JSON-formatted string; the “prototype chain” of the Book object is not preserved. On line 69, the restored array therefore contains only plain objects, not Book instances.

In this app, it happens that a plain object works just as well as a Book instance.
In general, however, a more appropriate approach is to add code that converts the restored array into an array of Book instances.

cjyuan (commented):
Changes look good. I have few more suggestions.

cjyuan:
Changes look good! Well done!

PR #302, opened 2025-08-13
LONDON | MAY 2025 | JESUS DEL MORAL | DATA FLOWS | BOOK LIBRARY

delmorallopez:
topics: debugging, DOM
My website should be able to:

View a list of books that I've read
Add books to a list of books that I've read
Including title, author, number of pages and if I've read it
If any of the information is missing it shouldn't add the book and should show an alert
Remove books from my list
Bugs to be fixed
Website loads but doesn't show any books
Error in console when you try to add a book
It uses the title name as the author name
Delete button is broken
When I add a book that I say I've read - it saves the wrong answer
I think there are other some other small bugs in my code...but I'm lazy so I can't fix them all.

I wish somebody would help me!

### DIFF
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -1,3 +1,5 @@
+
+
 let myLibrary = [];
 
 window.addEventListener("load", function (e) {
@@ -7,16 +9,15 @@ window.addEventListener("load", function (e) {
 
 function populateStorage() {
   if (myLibrary.length == 0) {
-    let book1 = new Book("Robison Crusoe", "Daniel Defoe", "252", true);
+    let book1 = new Book("Robison Crusoe", "Daniel Defoe", 252, true);
     let book2 = new Book(
       "The Old Man and the Sea",
       "Ernest Hemingway",
-      "127",
+      127,
       true
     );
     myLibrary.push(book1);
     myLibrary.push(book2);
-    render();
   }
 }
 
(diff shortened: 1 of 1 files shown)


delmorallopez:
Hi @cjyuan 
I went through all the debugging feadback points already, please let me know if its missing something 

Thank you

debugging/book-library/script.js
```diff
@@ -28,17 +29,19 @@ const check = document.getElementById("check");
...
+  const cleanAuthor = author.value.trim();
+  const cleanPages = parseInt(pages.value, 10);
+
+  if (!cleanTitle|| !cleanAuthor) {  // title and author doesn't allowed to contain only space characters
+    alert("Title and Author cannot be empty or spaces only."); 
+    return false;
+  } else if (isNaN(cleanPages) || cleanPages <= 0) {   // the value of type of pages must be a integer 
+    alert("Pages must be a positive whole number.");
     return false;
   } else {
```
cjyuan, line 42: When an `if` block contains a return statement, we could choose not to use `else`.

debugging/book-library/script.js
```diff
@@ -28,17 +29,19 @@ const check = document.getElementById("check");
...
-    let book = new Book(title.value, title.value, pages.value, check.checked);
-    library.push(book);
+    let book = new Book(cleanTitle, cleanAuthor, pages.value, check.checked);  //// title and author be allowed to contain leading or trailing space characters, we storage using parseInt
```
cjyuan, line 43: `pages.value` could be "3e1" (which could pass the check on line 39).

debugging/book-library/script.js
```diff
@@ -52,11 +55,10 @@ function Book(title, author, pages, check) {
 
 function render() {
   let table = document.getElementById("display");
-  let rowsNumber = table.rows.length;
   //delete old table
-  for (let n = rowsNumber - 1; n > 0; n-- {
-    table.deleteRow(n);
-  }
+  // Keep only the header row
+  table.innerHTML = table.rows[0].outerHTML;
```
cjyuan, line 60: Since the `<table>` also contains a `<tbody>` child element. The statement on line 60 will remove the `<tbody>` element.

An alternative approach is the clear `<tbody>` directly.

cjyuan (commented):

debugging/book-library/script.js
delmorallopez: @cjyuan 
I fixed all the errors

cjyuan (commented):
Changes look good.

PR #308, opened 2025-08-16
ZA Cape Town | ITP-May-2025 | Dawud Vermeulen | Sprint 2 | Book-Library

Dave-Vermeulen:
## Learners, PR Template

Self checklist

## Changelist

Fixed the errors in the code. some were small some were bigger like the submit() function had problems with author being set to title so changed to author value. Then went on to the render() function and fixed the delButton use. A 3rd bug was on the delete buttons event listener there was a typo. Then finally the if statement logic in the render function was reversed so fixed that.

## Questions

Please review my PR 🙏🏽

### DIFF
--- a/debugging/book-library/index.html
+++ b/debugging/book-library/index.html
@@ -31,7 +31,7 @@
       <div class="form-group">
         <label for="title">Title:</label>
         <input
-          type="title"
+          type="text"
           class="form-control"
           id="title"
           name="title"
@@ -39,7 +39,7 @@
         />
         <label for="author">Author: </label>
         <input
-          type="author"
+          type="text"
           class="form-control"
           id="author"
           name="author"
--- a/debugging/book-library/index.html
+++ b/debugging/book-library/index.html
@@ -22,20 +22,22 @@
     </button>
 
     <div id="demo" class="collapse">
-      <div class="form-group">
-        <label for="title">Title:</label>
-        <input type="text" class="form-control" id="title" name="title" required>
-        <label for="author">Author: </label>
-        <input type="text" class="form-control" id="author" name="author" required>
-        <label for="pages">Pages:</label>
-        <input type="number" class="form-control" id="pages" name="pages" required>
-        <label class="form-check-label">
-          <input type="checkbox" class="form-check-input" id="check" value="">Read
-        </label>
-        <input type="submit" value="Submit" class="btn btn-primary" onclick="submit();">
-      </div>
+      <form id="addBookForm">
+        <div class="form-group">
+          <label for="title">Title:</label>
+          <input type="text" class="form-control" id="title" name="title" required />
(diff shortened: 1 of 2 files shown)


cjyuan:
- Can you check if any of this general feedback can help you further improve your code?

- I think the PR title is not quite in the expected format.

debugging/book-library/index.html
LonMcGregor, line 1: There are some things missing from your HTML file that should be added. Hint: how people are going to interact with your webpage in a web browser?

There is also some invalid HTML. Can you spot these? Hint: have you used an HTML validator before? https://validator.w3.org/

debugging/book-library/script.js
```diff
@@ -1,103 +1,85 @@
...
-
 //check the right input from forms and if its ok -> add the new book (object in array)
 //via Book function and start render function
 function submit() {
   if (
-    title.value == null ||
-    title.value == "" ||
-    pages.value == null ||
-    pages.value == ""
+    titleInput.value.trim == "" || // cant be null so removed all that
```
LonMcGregor, line 21: There may be ways to improve your form validation here. Things to think about:

* Is your validation checking all the inputs?
* Is there a way to get the browser to do some of the checking for you, without needing to manually check for empty inputs?

debugging/book-library/script.js
```diff
@@ -1,103 +1,85 @@
...
+    const deleteCell = row.insertCell(4);
+    const deleteBtn = document.createElement("button");
+    deleteBtn.className = "btn btn-warning";
+    deleteBtn.innerHTML = "Delete";
+    deleteCell.appendChild(deleteBtn);
+    deleteBtn.addEventListener("click", function () {
+      const deletedTitle = myLibrary[index].title;
+      myLibrary.splice(index, 1);
       render();
+      alert(`You've deleted title: ${deletedTitle}`);
```
LonMcGregor, line 77: Think about how deleting a file works on your desktop. Is being notified after the deletion normally how this works? Is there a better interaction you could use than an `alert` here?

LonMcGregor (commented):
Hi, I've left some comments that should help improve your solution further. Well done on getting this far, you're close to being finished

debugging/book-library/script.js
Dave-Vermeulen: okay so this is bit _higher grade_. i think what you mean is that the users flow is interrupted when the alert pops up (clarify if im wrong kanala). based on this assumption and the fact that the delete works i have built a new function to display the notification and done away with the disruptive alert.

LonMcGregor: The alert is quite interrupting, yes, making it less "in your face" is one solution. The other is, if you really want something interrupting, to use something like `confirm` instead. 

It sounds like you've made good progress here. I can't see the files yet on github though - have you comitted them all?

PR #315, opened 2025-08-19
WM | ITP-MAY-25 | NAHOM MESFIN | Sprint 2 | Book Library

nahommesfin77:
## Learners, PR Template

Self checklist

## Changelist

I debugged and fixed issues preventing books from displaying. and corrected errors when adding a book. then fixed the title/author mix-up, ensured the delete button works, and corrected the read-status logic. I then implemented validation so incomplete book entries show an alert instead of being added to the list.

### DIFF
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -21,12 +21,6 @@ const check = document.getElementById("check");
 // Notification function
 function showNotification(message) {
   let notification = document.getElementById("notification");
-  if (!notification) {
-    notification = document.createElement("div");
-    notification.id = "notification";
-    notification.className = "notification";
-    document.body.insertBefore(notification, document.body.firstChild);
-  }
   notification.textContent = message;
   notification.style.display = "block";
   setTimeout(() => {
@@ -34,29 +28,25 @@ function showNotification(message) {
   }, 3000); // 3 seconds
 }
 
+// Listen for form submission
+document.getElementById("bookForm").addEventListener("submit", function(event) {
+  event.preventDefault();
+  submit();
+});
+
 function submit() {
   const titleValue = title.value.trim();
   const authorValue = author.value.trim();
   const pagesValue = pages.value.trim();
 
-  if (!titleValue || !authorValue || !pagesValue) {
-    alert("Please fill all fields!");
-    return false;
-  }
-
-  // Prevent numbers in Author field
-  if (!/^[a-zA-Z\s.'-]+$/.test(authorValue)) {
-    alert("Author name can only contain letters, spaces, apostrophes, periods, or hyphens.");
-    return false;
-  }
-  // pages must be a positive number
+  // Additional validation for pages
   const pagesNumber = Number(pagesValue);
   if (!Number.isInteger(pagesNumber) || pagesNumber <= 0) {
-    alert("Number of pages must be a positive whole number.");
-    return false;
+    showNotification("Number of pages must be a positive whole number.");
+    return;
   }
 
-  let book = new Book(titleValue, authorValue, Number(pagesValue), check.checked);
+  let book = new Book(titleValue, authorValue, pagesNumber, check.checked);
   myLibrary.push(book);
   render();
 


cjyuan (commented):

debugging/book-library/script.js
```diff
@@ -25,22 +18,31 @@ const author = document.getElementById("author");
...
-    title.value == null ||
-    title.value == "" ||
-    pages.value == null ||
-    pages.value == ""
-  ) {
+  const titleValue = title.value.trim();
+  const authorValue = author.value.trim();
+  const pagesValue = pages.value.trim();
+
+  if (!titleValue || !authorValue || !pagesValue) {
```
LonMcGregor, line 26: Are there any values in the form you might want to validate more carefully than just "is there data here"?

debugging/book-library/script.js
```diff
@@ -51,53 +53,47 @@ function Book(title, author, pages, check) {
...
+    const deleteCell = row.insertCell(4);
+    const deleteButton = document.createElement("button");
+    deleteButton.className = "btn btn-warning";
+    deleteButton.textContent = "Delete";
+    deleteCell.appendChild(deleteButton);
+
+    deleteButton.addEventListener("click", () => {
+      myLibrary.splice(i, 1);  // delete immediately
+      render();                 // update the table
+      alert(`Deleted "${book.title}" successfully.`); // optional confirmation
```
LonMcGregor, line 91: When you are deleting something like a file on your PC, do you normally expect to be notified after, or is there an interaction that would work that is better than `alert`?

debugging/book-library/index.html
```diff
@@ -31,15 +29,15 @@ <h1>Library</h1>
       <div class="form-group">
         <label for="title">Title:</label>
         <input
-          type="title"
+          type="text"
```
LonMcGregor, line 32: Right now you need to manually validate the form. Do you know what change you need to make to the HTML and script to make the browser validate the form inputs automatically?

LonMcGregor (commented):
Good work on this sprint. It is almost done. I have left some remarks for how you could improve it further

debugging/book-library/script.js
nahommesfin77: yes there are a couple of validation rules i should have added.

> line 32

I added a conditon `if (!/^[a-zA-Z\s.'-]+$/.test(authorValue)` to validate authors value, limiting it to only letters, spaces, periods, apostrophes, and hyphens.

> line 38

I added another condition for page number `if (!Number.isInteger(pagesNumber) || pagesNumber <= 0)` to validate only postive numbers.

debugging/book-library/index.html
```diff
@@ -28,51 +27,53 @@ <h1>Library</h1>
...
         <label for="author">Author: </label>
         <input
-          type="author"
+          type="text"
           class="form-control"
           id="author"
           name="author"
           required
+          pattern="[a-zA-Z\s.'-]+"
+          title="Only letters, spaces, apostrophes, periods, or hyphens allowed"
```
cjyuan, line 48: This approach is good but the pattern might be too restrictive.

debugging/book-library/script.js
```diff
@@ -25,22 +18,43 @@ const author = document.getElementById("author");
...
+  setTimeout(() => {
+    notification.style.display = "none";
+  }, 3000); // 3 seconds
+}
+
+// Listen for form submission
+document.getElementById("bookForm").addEventListener("submit", function(event) {
+  event.preventDefault();
+  submit();
+});
```
cjyuan, line 35: Placing code that's to be executed once on page load (lines 3-6, 32-35) together could make locating this kind of code easier.

debugging/book-library/script.js
```diff
@@ -25,22 +18,43 @@ const author = document.getElementById("author");
...
+
+  let book = new Book(titleValue, authorValue, pagesNumber, check.checked);
+  myLibrary.push(book);
+  render();
+
+  // Reset form
+  title.value = "";
+  author.value = "";
+  pages.value = "";
+  check.checked = false;
```
cjyuan, line 57: We could also use `.reset()`  of a form element to clear the form.

debugging/book-library/script.js
```diff
@@ -51,53 +65,47 @@ function Book(title, author, pages, check) {
...
-    delBut.id = i + 5;
-    deleteCell.appendChild(delBut);
-    delBut.className = "btn btn-warning";
-    delBut.innerHTML = "Delete";
-    delBut.addEventListener("clicks", function () {
-      alert(`You've deleted title: ${myLibrary[i].title}`);
-      myLibrary.splice(i, 1);
-      render();
+    // Delete button
+    const deleteCell = row.insertCell(4);
```
cjyuan, line 94: Can you think of the pros and cons of these two approaches for creating cells within a row?
- Keeping all the cell creation code in one location, like the original code does.
- Scattering the cell creation code across different locations, like what you did.

cjyuan (commented):
The changes you made look good. You can go through my additional suggestions later when you have time.

PR #344, opened 2025-12-07
London | 25-ITP-Sep | Adnaan Abo | Sprint 2 | book library project

AdnaanA:
Self checklist

## Changelist

1. Website loads but doesn't show any books. fixed
2. Error in console when you try to add a book. fixed
3. It uses the title name as the author name. fixed
4. Delete button is broken. fied
5. When I add a book that I say I've read - it saves the wrong answer. fixed

also ran html validation using w3c and fixed all the recommendations.

### DIFF
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -61,19 +61,25 @@ function Book(titleValue, authorValue, pagesNumber, isRead) {
 
 function render() {
   const table = document.getElementById("display");
+  const tbody = table.querySelector("tbody");
+  tbody.innerHTML = ""; // clear previous rows
 
-  const tbody = table.querySelector('tbody');
-  tbody.innerHTML = ''; // clears all rows efficiently
-  
   myLibrary.forEach((book, i) => {
-    const row = tbody.insertRow(); // insert row into tbody
+    const row = tbody.insertRow();
 
-    row.insertCell(0).textContent = book.bookTitle;
-    row.insertCell(1).textContent = book.author;
-    row.insertCell(2).textContent = book.pages;
-
-    // read/unread toggle
+    // Create ALL CELLS first (like original code)
+    const titleCell = row.insertCell(0);
+    const authorCell = row.insertCell(1);
+    const pagesCell = row.insertCell(2);
     const wasReadCell = row.insertCell(3);
+    const deleteCell = row.insertCell(4);
+
+    // Fill text cells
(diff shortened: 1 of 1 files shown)


debugging/book-library/script.js
```diff
@@ -1,103 +1,89 @@
...
-    let book = new Book(title.value, title.value, pages.value, check.checked);
-    library.push(book);
-    render();
+    return;
   }
+
+  const book = new Book(
+    bookTitle.value.trim(),
+    author.value.trim(),
+    pages.value.trim(),
```
cjyuan: - It is a good practice to save sanitized or pre-processed values in some variables first
  - to avoid calling `.trim()` repeatedly, and 
  - to reduce the chance of accidently using the raw value.

- The book's `page` property can still be assigned some invalid page count.

debugging/book-library/script.js
```diff
@@ -1,103 +1,89 @@
...
+      book.check = !book.check;
       render();
     });
 
-    //add delete button to every row and render again
-    let delButton = document.createElement("button");
-    delBut.id = i + 5;
-    deleteCell.appendChild(delBut);
+    // delete button
+    const deleteCell = row.insertCell(4);
```
cjyuan: Can you think of the pros and cons of these two approaches for creating cells within a row?
- Keeping all the cell creation code in one location, like the original code does.
- Scattering the cell creation code across different locations, like what you did.

debugging/book-library/script.js
```diff
@@ -1,103 +1,89 @@
...
-  let rowsNumber = table.rows.length;
-  //delete old table
-  for (let n = rowsNumber - 1; n > 0; n-- {
-    table.deleteRow(n);
+  const table = document.getElementById("display");
+
+  // remove all rows except the header
+  while (table.rows.length > 1) {
+    table.deleteRow(1);
   }
```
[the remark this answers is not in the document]
AdnaanA: we can use  tbody.innerHTML = ''; to clear all rows efficiently.

however we have to use const row = tbody.insertRow(); to insert rows into tbody.

debugging/book-library/script.js
AdnaanA: Keeping all cells created in one block is the best practice.
It is clearer, safer, easier to maintain, and less likely to produce bugs.
while
Scattered cell creation should only be used when absolutely necessary for conditional or dynamic table structures.

PR #345, opened 2025-12-07
Glasgow | 25-ITP-SEP | Fares Bakhet | Sprint 2 | Book-library

Fares-Bakhet:
Learners, PR Template

Self checklist

## Changelist

Complete the book library adjustments.

## Questions

No questions.

### DIFF
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -7,12 +7,10 @@ class Book {
   }
 }
 
-
 const library = [
   new Book("Robinson Crusoe", "Daniel Defoe", 252, true),
   new Book("The Old Man and the Sea", "Ernest Hemingway", 127, true),
 ];
-
 
 const bookForm = document.getElementById("bookForm");
 const titleInput = document.getElementById("titleInput");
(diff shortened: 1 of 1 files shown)


debugging/book-library/script.js
```diff
@@ -1,103 +1,109 @@
...
+function render() {
+
+  tableBody.innerHTML = "";
+
+  library.forEach((book, index) => {
+    const row = document.createElement("tr");
+
+    row.innerHTML = `
+      <td>${book.title}</td>
+      <td>${book.author}</td>
```
cjyuan: Can you suggest an approach to combat special characters such as '<' and '&' that may exist in a book title? If these characters are not handled properly, they may ruin the display.

Fares-Bakhet: @cjyuan, Okay, will use "textContent" instead.

cjyuan: You could also write a function to escape the special characters in the string. (Just ask AI. The function is just a few lines of code).

Then you would be able to express something like:

```
      `...
      <td>${escape(book.title)}</td>
      <td>${escape(book.author)}</td>
      ...
      `
```

This way, you can keep using your approach to construct the elements.

PR #348, opened 2025-12-08
Glasgow | 25-ITP-Sep |Hanna Mykytiuk| Sprint 2 | Book library

HannaOdud:
## Learners, PR Template

Self checklist

## Changelist

Bugfixing of book library app.

## Questions

No questions

### DIFF
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -31,12 +31,21 @@ document.getElementById('submit-btn').addEventListener('click', submit);
 function submit() {
   const title = sanitize(titleDom.value.trim());
   const author = sanitize(authorDom.value.trim());
-  const pages = sanitize(pagesDom.value.trim());
+  let pages = sanitize(pagesDom.value.trim());
+  if (isNaN(pages)){
+    alert("Enter valid value for page number!");
+    return false;
+  }
+  pages = parseInt(pages);
+  if (!isValueInteger(pages)){
+    alert("Enter valid value for page number!");
+    return false;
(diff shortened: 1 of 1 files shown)


debugging/book-library/script.js
```diff
@@ -20,25 +19,26 @@ function populateStorage() {
...
 //via Book function and start render function
 function submit() {
   if (
-    title.value == null ||
-    title.value == "" ||
-    pages.value == null ||
-    pages.value == ""
+    titleDom.value == "" || 
+    authorDom.value == "" ||
+    pagesDom.value == 0
```
cjyuan: What is the type of the `pagesDom.value`? Why compare it to `0`?

debugging/book-library/script.js
```diff
@@ -20,25 +19,26 @@ function populateStorage() {
...
   ) {
     alert("Please fill all fields!");
     return false;
   } else {
-    let book = new Book(title.value, title.value, pages.value, check.checked);
-    library.push(book);
+    let book = new Book(titleDom.value.trim(), authorDom.value.trim(), pagesDom.value, checkDom.checked);
```
cjyuan: The values you used on line 40 are not the values you checked on lines 33-35. A book may end up having an empty string as its title or a negative number as its page count.

A better practice is 
Step 1: Sanitised and preprocessed the raw input values and stored the resulting value in some variables.
Step 2; Check the values in the variables
Step 3: If the values are ok, use the values stored in the variables.

PR #385, opened 2026-03-03
London | 26-ITP-Jan | Abdul Moiz | Sprint 2 | Book Library

A-Moiz:
## Learners, PR Template

Self checklist

## Changelist

- Resolved all known bugs
- Reset all values after adding book to improve UX

## Questions

N/A

### DIFF
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -1,3 +1,42 @@
+// Book class
+class Book {
+  constructor(title, author, pages, check) {
+    const validated = constructorValidation(title, author, pages, check);
+
+    this.title = validated.title;
+    this.author = validated.author;
+    this.pages = validated.pages;
+    this.check = validated.check;
+  }
+}
+
+// Checking if constructor arguments are valid
+function constructorValidation(title, author, pages, check) {
+  if (typeof title !== "string" || title.trim() === "") {
+    throw new Error("Title must be a non-empty string");
+  }
+
+  if (typeof author !== "string" || author.trim() === "") {
+    throw new Error("Author must be a non-empty string");
+  }
+
+  const pagesNum = Number(pages);
+  if (isNaN(pagesNum) || pagesNum <= 0) {
+    throw new Error("Pages must be a positive number");
+  }
+
+  if (typeof check !== "boolean") {
+    throw new Error("Check must be a boolean");
+  }
+
+  return {
+    title: title.trim(),
+    author: author.trim(),
+    pages: pagesNum,
+    check,
+  };
+}
+
 let myLibrary = [];
 
 window.addEventListener("load", function (e) {
@@ -7,8 +46,8 @@ window.addEventListener("load", function (e) {
 
 function populateStorage() {
   if (myLibrary.length === 0) {
-    let book1 = new Book("Robison Crusoe", "Daniel Defoe", "252", true);
-    let book2 = new Book(
+    const book1 = new Book("Robison Crusoe", "Daniel Defoe", "252", true);
+    const book2 = new Book(
       "The Old Man and the Sea",
       "Ernest Hemingway",
       "127",
@@ -16,7 +55,6 @@ function populateStorage() {
     );
     myLibrary.push(book1);
     myLibrary.push(book2);
-    render();
   }
 }
 
@@ -32,27 +70,13 @@ function submit() {
     alert("Please fill all fields!");
     return false;
   } else {
-    let book = new Book(
-      title.value,
-      author.value,
-      String(pages.value),
-      check.checked
-    );
+    let book = new Book(title.value, author.value, pages.value, check.checked);
     myLibrary.push(book);
     render();
     title.value = "";
     author.value = "";
     pages.value = "";
     check.checked = false;
-  }
-}
-
-class Book {
-  constructor(title, author, pages, check) {
-    this.title = title;
-    this.author = author;
-    this.pages = pages;
-    this.check = check;
   }
 }
 
@@ -82,11 +106,7 @@ function render() {
     changeBut.className = "btn btn-success";
     wasReadCell.appendChild(changeBut);
     let readStatus = "";
-    if (myLibrary[i].check === false) {
-      readStatus = "No";
-    } else {
-      readStatus = "Yes";
-    }
+    myLibrary[i].check ? (readStatus = "Yes") : (readStatus = "No");
     changeBut.innerText = readStatus;
 
     changeBut.addEventListener("click", function () {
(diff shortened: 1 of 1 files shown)


debugging/book-library/script.js
```diff
@@ -6,7 +6,7 @@ window.addEventListener("load", function (e) {
 });
 
 function populateStorage() {
-  if (myLibrary.length == 0) {
+  if (myLibrary.length === 0) {
     let book1 = new Book("Robison Crusoe", "Daniel Defoe", "252", true);
```
ykamal: Love to see some OOP code. Since the class you're referencing here is declared further down the file, for the sake of readability and code organisation, I recommend moving the class definition to much higher, or to the beginning of this file.

ykamal: `book1` is not getting reassigned later, so using a `const` keyword will make it safer in the same scope. It's still going to allow you to run object mutations where applicable. but it ensures that you cannot redeclare a variable with the same name inside the same scope.
Same for `book2`

debugging/book-library/script.js
```diff
@@ -28,33 +28,39 @@ const check = document.getElementById("check");
 //check the right input from forms and if its ok -> add the new book (object in array)
 //via Book function and start render function
 function submit() {
-  if (
-    title.value == null ||
-    title.value == "" ||
-    pages.value == null ||
-    pages.value == ""
-  ) {
+  if (title.value === "" || author.value === "" || pages.value === "") {
```
ykamal: Good validation check 👍 
However, if you do not remove unnecessary whitespaces from the start and the end, `"   "` will pass through this check as it's a bunch of valid characters and is not an empty string. How do you think you could counter this?

debugging/book-library/script.js
```diff
@@ -28,33 +28,39 @@ const check = document.getElementById("check");
...
     alert("Please fill all fields!");
     return false;
   } else {
-    let book = new Book(title.value, title.value, pages.value, check.checked);
-    library.push(book);
+    let book = new Book(
+      title.value,
+      author.value,
+      String(pages.value),
```
ykamal: This is good safe practice, but there is most likely no need to convert this into a string, as all input values are always returned (usually) as strings, unless something like this is used: https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/valueAsNumber

debugging/book-library/script.js
```diff
@@ -28,33 +28,39 @@ const check = document.getElementById("check");
...
-  this.title = title;
-  this.author = author;
-  this.pages = pages;
-  this.check = check;
+class Book {
+  constructor(title, author, pages, check) {
+    this.title = title;
+    this.author = author;
+    this.pages = pages;
+    this.check = check;
```
ykamal: Since this is JavaScript, do you think we should ensure that these constructor arguments are provided in the right data type?

ykamal: + and data range (numbers)

debugging/book-library/script.js
ykamal: `render()` appears to be called twice on page load. Perhaps you should create a sequence of this from the root function call and not inside here?

debugging/book-library/script.js
```diff
@@ -76,10 +82,10 @@ function render() {
...
     let readStatus = "";
-    if (myLibrary[i].check == false) {
-      readStatus = "Yes";
-    } else {
+    if (myLibrary[i].check === false) {
       readStatus = "No";
+    } else {
+      readStatus = "Yes";
     }
     changeBut.innerText = readStatus;
```
ykamal, line 114: Have you come across Ternary Operator? Could make the code cleaner:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_operator

debugging/book-library/script.js
```diff
@@ -90,11 +96,11 @@ function render() {
 
     //add delete button to every row and render again
     let delButton = document.createElement("button");
-    delBut.id = i + 5;
-    deleteCell.appendChild(delBut);
-    delBut.className = "btn btn-warning";
-    delBut.innerHTML = "Delete";
-    delBut.addEventListener("clicks", function () {
+    delButton.id = i + 5;
```
ykamal: I can see that you are trying to avoid duplicate IDs. Nice approach. However, the approach is hard-coding a specific number, and the outcome could be inconsistent.
You could just use `delButton.id = "delButton_" + i` or such tactics. That'll make them truly unique. Obviously, use string interpolation with backticks. I can't write that here.

debugging/book-library/script.js
```diff
@@ -90,11 +96,11 @@ function render() {
...
+    deleteCell.appendChild(delButton);
+    delButton.className = "btn btn-warning";
+    delButton.innerHTML = "Delete";
```
ykamal: use `textContent` as it parses the provided value as a string, whereas `innerHTML` parses it as HTML and should only be used if the change contains actual HTML. You can also use `innerText` for static text. `textContent` may be faster for most things.

debugging/book-library/script.js
```diff
@@ -90,11 +96,11 @@ function render() {
...
+    delButton.addEventListener("click", function (e) {
```
ykamal, line 129: a better way to do this would be using `data-` attributes. You're using `i` inside a closure which, while fine, is not the best practice. 
jQuery has excellent APIs for using data attributes. Even with core JavaScript, if you can master `data` attributes, you will no longer need to do these, and can use HTML itself for storing that reference.
https://developer.mozilla.org/en-US/docs/Web/HTML/How_to/Use_data_attributes

ykamal (changes requested):
Nice work! 👍 Things look mostly right. Well done. There are a few things that need a bit of work. I've left comments below. Let me know if you have any questions.

A-Moiz:
Hi @ykamal , I've made all the changes you requested. Let me know if you anymore recommendations.

ykamal:
Apologies, I have not been well. Going to look now.

debugging/book-library/script.js
ykamal: @A-Moiz there's an easier way to do this:
```
const myBool = anotherBool ? "Yes" : "No";
```

debugging/book-library/script.js
ykamal: I don't think this has been adressed

debugging/book-library/script.js
ykamal: This is much more appropriate as this isn't dealing with stale data inside a closure

ykamal (changes requested):

A-Moiz:
Hi @ykamal , thanks for your comments! I've made the changes. Please have another look and let me know what you think.

cjyuan:
@A-Moiz  It seems your changes have addressed @ykamal comments, I will mark this PR as "Complete" on behalf of the reviewer.

illicitonion:

PR #398, opened 2026-04-04
London | 26-ITP-January | Eugenie Ahangama | Sprint 3 | Programmer Humor

Eugenie-A:
## Learners, PR Template

Self checklist

## Changelist

- Created HTML, CSS and JavaScript files
- Fetches JSON using `fetch`
- Function makes API call to the correct endpoint
- Logs received data to the console
- Renders the `img` property into an `` tag in the DOM
- Error handling with `try/catch`

## Questions

Is there a better way to handle errors than just console.log(error) - should I be displaying something to the user instead?

### DIFF
--- a/fetch/programmer-humour/script.js
+++ b/fetch/programmer-humour/script.js
@@ -1,0 +1,25 @@
+async function getComic() {
+  try {
+    // Fetch the latest comic from the API
+    const response = await fetch("https://xkcd.now.sh/?comic=latest");
+    // Convert the response to JSON
+    const data = await response.json();
+    // Log the data to the console
+    console.log(data);
+
+    // Grab the comic container from the DOM
+    let comicDiv = document.getElementById("comic");
+    // Create an img element
+    let imgComic = document.createElement("img");
+    // Set the src of the img to the comic image URL
+    imgComic.src = data.img;
+    // Append the img to the comic container
+    comicDiv.appendChild(imgComic);
+  } catch (error) {
+    // Catch and log any errors
+    console.log(error);
+  }
+}
+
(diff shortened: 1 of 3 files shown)


fetch/programmer-humour/script.js
```diff
@@ -0,0 +1,25 @@
...
+    // Convert the response to JSON
+    const data = await response.json();
+    // Log the data to the console
+    console.log(data);
+
+    // Grab the comic container from the DOM
+    let comicDiv = document.getElementById("comic");
+    // Create an img element
+    let imgComic = document.createElement("img");
+    // Set the src of the img to the comic image URL
```
Luro91, line 14: I understand that comments can help you writing the code. It is best practice to only place comments where they are needed for understanding. The code should be self explanatory and comments should mainly explain why things are written in a specific way. 

From https://mitcommlab.mit.edu/broad/commkit/coding-and-comment-style/
> The most important concept here is that the code itself is the primary means of commenting and documentation. 
> Use comments only to add context or explain choices that cannot be expressed through thoughtful naming, structure or the context of operations.

You could write the comments during the development and then remove them before committing it. Try to use less of them over time

Luro91 (commented):
Well done. This is a nice implementation and you used meaningful commit messages that make it easy to understand what each commit does.

Regarding your question:
> Is there a better way to handle errors than just console.log(error) - should I be displaying something to the user instead?

In general it's good practice in terms of UX to give the user feedback if an error occurs. This can be done by showing an error message. (As an advanced topic you could also show a loader animation while the data is being fetched)

illicitonion:

PR #413, opened 2026-04-09
Sheffield | 26-Jan-ITP | Martha Ogunbiyi | Sprint 1 | Object Destructuring

marthak1:
## Learners, PR Template

Self checklist

## Changelist

Sprint 1 Exercise - Object destructing complete

### DIFF
--- a/Sprint-1/destructuring/exercise-1/exercise.js
+++ b/Sprint-1/destructuring/exercise-1/exercise.js
@@ -3,10 +3,10 @@ const personOne = {
   age: 34,
   favouriteFood: "Spinach",
 };
-
+let {name, age, favouriteFood} = personOne
 // Update the parameter to this function to make it work.
 // Don't change anything else.
-function introduceYourself(___________________________) {
+function introduceYourself(aboutMe){
   console.log(
     `Hello, my name is ${name}. I am ${age} years old and my favourite food is ${favouriteFood}.`
   );
--- a/Sprint-1/destructuring/exercise-2/exercise.js
+++ b/Sprint-1/destructuring/exercise-2/exercise.js
@@ -70,10 +70,20 @@ let hogwarts = [
     occupation: "Teacher",
   },
 ];
-function gryffindors(hogwartsArray){
-  let houseOfGryffindor = hogwartsArray.filter( (item) => item.house === "Gryffindor");
-   houseOfGryffindor.forEach( ({firstName, lastName}) => {
-console.log(`${firstName} ${lastName}`);
-   })
+function gryffindors(hogwartsArray) {
+  let houseOfGryffindor = hogwartsArray.filter(
+    (item) => item.house === "Gryffindor"
+  );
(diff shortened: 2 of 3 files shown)


Sprint-1/destructuring/exercise-1/exercise.js
```diff
@@ -3,10 +3,10 @@ const personOne = {
   age: 34,
   favouriteFood: "Spinach",
 };
-
+let {name, age, favouriteFood} = personOne
```
Luro91, line 6: The instructions say to only change the parameter of the function and nothing else

Sprint-1/destructuring/exercise-2/exercise.js
```diff
@@ -70,3 +70,20 @@ let hogwarts = [
...
+  });
+}
+gryffindors(hogwarts);
+
+function teachersWithPets(hogwartsArray) {
+  let namesOfTeachers = hogwartsArray.filter((item) => item.pet !== null && item.occupation === "Teacher");
+  namesOfTeachers.forEach(({ firstName, lastName }) => {
+    console.log(`${firstName} ${lastName}`);
+  });
+}
```
Luro91: Combining filter and for each works correctly. 
Something to consider for bigger lists is performance. In that cases you could only loop through each list once and doing both operations inside one loop (checking the condition and logging)

Sprint-1/destructuring/exercise-3/exercise.js
```diff
@@ -6,3 +6,22 @@ let order = [
   { itemName: "Hot Coffee", quantity: 2, unitPricePence: 100 },
   { itemName: "Hash Brown", quantity: 4, unitPricePence: 40 },
 ];
+function orders(orderItems) {
+    const header = `${"QTY".padEnd(5)}${"ITEM".padEnd(20)}${"TOTAL (£)".padStart(10)}`;
+    console.log(header);
+  orderItems.forEach(({ quantity, itemName, unitPricePence }) => {
```
Luro91, line 14: The code formatting is not consistent. How could you ensure the code is formatted consistently automatically?

marthak1: Thanks for the feedback the code is now formatted with prettier

Sprint-1/destructuring/exercise-2/exercise.js
marthak1: Thanks for the suggestion. That makes sense—using filter and forEach results in iterating twice. I’ve updated the implementation to use a single loop, combining the condition check and logging to improve performance for larger datasets.

Sprint-1/destructuring/exercise-1/exercise.js
marthak1: Thanks for the feedback, this has now been updated to meet requirements.

Luro91: I still see `let {name, age, favouriteFood} = personOne` defined outside of the function

illicitonion:

PR #415, opened 2026-04-09
London | 26-ITP-Jan | Angela McLeary | Sprint 2 | Book Library

AngelaMcLeary:
## Learners, PR Template

Self checklist

## Changelist

Debugging of book library app and fixing typos, syntaxes and validation to mention a few.

### DIFF
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -38,7 +38,7 @@ function submit() {
     return false;
   } else {
     let book = new Book(title.value, title.value, pages.value, check.checked);
-    library.push(book);
+    myLibrary.push(book);
     render();
   }
 }
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -37,7 +37,7 @@ function submit() {
     alert("Please fill all fields!");
     return false;
   } else {
-    let book = new Book(title.value, title.value, pages.value, check.checked);
+    let book = new Book(title.value, author.value, pages.value, check.checked);
     myLibrary.push(book);
     render();
   }
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -54,7 +54,7 @@ function render() {
   let table = document.getElementById("display");
   let rowsNumber = table.rows.length;
   //delete old table
-  for (let n = rowsNumber - 1; n > 0; n-- {
+  for (let n = rowsNumber - 1; n > 0; n-- ){
     table.deleteRow(n);
   }
   //insert updated row and cells
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -77,9 +77,9 @@ function render() {
     wasReadCell.appendChild(changeBut);
     let readStatus = "";
     if (myLibrary[i].check == false) {
-      readStatus = "Yes";
-    } else {
       readStatus = "No";
+    } else {
+      readStatus = "Yes";
     }
     changeBut.innerText = readStatus;
 
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -89,7 +89,7 @@ function render() {
     });
 
     //add delete button to every row and render again
-    let delButton = document.createElement("button");
+    let delBut = document.createElement("button");
     delBut.id = i + 5;
     deleteCell.appendChild(delBut);
     delBut.className = "btn btn-warning";
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -81,30 +81,30 @@ function render() {
     pagesCell.innerHTML = myLibrary[i].pages;
 
     //add and wait for action for read/unread button
-    let changeBut = document.createElement("button");
-    changeBut.id = i;
-    changeBut.className = "btn btn-success";
-    wasReadCell.appendChild(changeBut);
+    let changeBtn = document.createElement("button");
+    changeBtn.id = i;
+    changeBtn.className = "btn btn-success";
+    wasReadCell.appendChild(changeBtn);
     let readStatus = "";
     if (myLibrary[i].check == false) {
       readStatus = "No";
     } else {
       readStatus = "Yes";
     }
-    changeBut.innerText = readStatus;
+    changeBtn.innerText = readStatus;
 
-    changeBut.addEventListener("click", function () {
+    changeBtn.addEventListener("click", function () {
       myLibrary[i].check = !myLibrary[i].check;
       render();
     });
 
     //add delete button to every row and render again
-    let delBut = document.createElement("button");
-    delBut.id = i + 5;
-    deleteCell.appendChild(delBut);
-    delBut.className = "btn btn-warning";
-    delBut.innerHTML = "Delete";
-    delBut.addEventListener("click", function (index) {
+    let deleteBtn = document.createElement("button");
+    deleteBtn.id = i + 5;
+    deleteCell.appendChild(delBtn);
+    deleteBtn.className = "btn btn-warning";
+    deleteBtn.innerHTML = "Delete";
+    delBtn.addEventListener("click", function (index) {
       alert(`You've deleted title: ${myLibrary[i].title}`);
       myLibrary.splice(i, 1);
       render();
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -79,9 +79,9 @@ function render() {
 
     //add and wait for action for read/unread button
     let changeBtn = document.createElement("button");
-    changeBtn.id = i;
     changeBtn.className = "btn btn-success";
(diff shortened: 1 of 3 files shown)


cjyuan (commented):

debugging/book-library/script.js
```diff
@@ -7,7 +7,7 @@ window.addEventListener("load", function (e) {
 
 function populateStorage() {
   if (myLibrary.length == 0) {
-    let book1 = new Book("Robison Crusoe", "Daniel Defoe", "252", true);
+    let book1 = new Book("Robinson Crusoe", "Daniel Defoe", "252", true);
```
cjyuan, line 10: Did you notice the third parameter is a string and not a number?

debugging/book-library/script.js
```diff
@@ -31,15 +31,25 @@ function submit() {
   if (
     title.value == null ||
     title.value == "" ||
+    author.value == null ||
```
cjyuan: Why compare `.value` to `null`? When will `.value` ever be `null`?

debugging/book-library/script.js
```diff
@@ -31,15 +31,25 @@ function submit() {
...
+    parseInt(pages.value) <= 0 ||
+    isNaN(parseInt(pages.value))
   ) {
-    alert("Please fill all fields!");
+    alert("Please fill all fields with valid input!");
     return false;
   } else {
-    let book = new Book(title.value, title.value, pages.value, check.checked);
-    library.push(book);
+    let book = new Book(title.value, author.value, pages.value, check.checked);
```
cjyuan: Using raw input value without properly pre-processing them is dangerous.

Here are some input that can "mess up" the app:
```
 title.value = "                       ";
 page.value = "3e-2";
```

debugging/book-library/script.js
cjyuan, line 1: Your latest commit changed `const` back to `let`.

It's a good practice to click the "File changes" in your own PR to review (double check) the code your submitted.

Image

debugging/book-library/script.js
```diff
@@ -71,30 +81,30 @@ function render() {
...
     let readStatus = "";
     if (myLibrary[i].check == false) {
-      readStatus = "Yes";
-    } else {
       readStatus = "No";
+    } else {
+      readStatus = "Yes";
     }
-    changeBut.innerText = readStatus;
+    changeBtn.innerText = readStatus;
```
cjyuan: This could be a good opportunity to practice using the `? :` conditional operator. Can you rewrite the code on lines 88–94 as a single statement?

debugging/book-library/script.js
```diff
@@ -71,30 +81,30 @@ function render() {
...
 
     //add delete button to every row and render again
-    let delButton = document.createElement("button");
-    delBut.id = i + 5;
-    deleteCell.appendChild(delBut);
-    delBut.className = "btn btn-warning";
-    delBut.innerHTML = "Delete";
-    delBut.addEventListener("clicks", function () {
+    let deleteBtn = document.createElement("button");
+    deleteBtn.id = i + 5;
```
cjyuan: - Lines 85, 103:

debugging/book-library/script.js
```diff
@@ -71,30 +81,30 @@ function render() {
...
+    deleteCell.appendChild(deleteBtn);
+    deleteBtn.className = "btn btn-warning";
+    deleteBtn.innerHTML = "Delete";
+    deleteBtn.addEventListener("click", function (index) {
```
cjyuan: Why introduce the parameter `index` to the callback function?

debugging/book-library/script.js
```diff
@@ -71,30 +81,30 @@ function render() {
...
       alert(`You've deleted title: ${myLibrary[i].title}`);
       myLibrary.splice(i, 1);
       render();
```

debugging/book-library/script.js
AngelaMcLeary: Hi @cjyuan, Thanks for your feedback. I did not notice that, thanks for pointing it out. Since the input is a string, I have converted the pages to numbers.

debugging/book-library/script.js
AngelaMcLeary: Hi @cjyuan, Thank you for your feedback. You are right, .value is never null. It returns a string or "" when empty. So that check was unnecessary. I have also refactored that code block and removed the unnecessary checks.

debugging/book-library/script.js
AngelaMcLeary: Hi @cjyuan, Thank you for your feedback. Good suggestion — I’ve replaced the if statement with a ternary operator for a cleaner single-line expression.

debugging/book-library/script.js
AngelaMcLeary: Hi @cjyuan, Thank you for your feedback. Good point - the id attributes were based on the loop index, so they are not guaranteed to be unique after deletions. Also since I'm not using them for DOM selection, so they are unnecessary, so I've removed both (changeBtn.id and delete.id).

debugging/book-library/script.js
AngelaMcLeary: Hi @cjyuan, Thank you for your feedback. That is definitely a mistake on my part. Furthermore the callback doesn't receive a custom index, only the event object. I wasn't even using it, I've deleted it. I have also refactored the code.

debugging/book-library/script.js
AngelaMcLeary: Hi @cjyuan, Thank you for your feedback. I have updated the logic so the book title is stored before calling splice, and the alert now runs after the deletion to ensure the correct title is shown.

debugging/book-library/script.js
AngelaMcLeary: Hi @cjyuan, Thank you for the feedback. I will double check in future.

debugging/book-library/script.js
```diff
@@ -29,32 +29,52 @@ const check = document.getElementById("check");
...
 }
 
 function Book(title, author, pages, check) {
   this.title = title;
   this.author = author;
-  this.pages = pages;
+  this.pages = Number(pages);
   this.check = check;
 }
+function alertDeleteToast(message) {
```
AngelaMcLeary, line 77: Hi @cjyuan, I replaced alert() with a toast notification to avoid blocking execution. The toast displays feedback after the delete operation completes and supports styled, multi-line messages for better readability and user experience.

debugging/book-library/script.js
```diff
@@ -71,33 +91,31 @@ function render() {
...
+    deleteBtn.className = "btn btn-warning";
+    deleteBtn.innerHTML = "Delete";
+    deleteCell.appendChild(deleteBtn);
+
+    deleteBtn.addEventListener("click", function () {
+      const deleteTitle = myLibrary[i].title;
       myLibrary.splice(i, 1);
       render();
+      alertDeleteToast(
+        `You've deleted title: <br> <strong>${deleteTitle}</strong>`
```
cjyuan: Note: Since this argument is to be assigned to `.innerHTML` of an element, best practice is to first escape
all special characters in `deleteTitle` to prevent corruption of the display or HTML injection.

(No change needed)

debugging/book-library/script.js
cjyuan: Can you address this comment? Currently some input values can still "mess up" the app.

AngelaMcLeary: Hi @cjyuan, Thank you for your feed back.  I’ve updated the validation. I now trim all input values before checking them, and I added a stricter regex (/^[1-9]\d*$/) for the pages field so it only accepts whole positive numbers without leading zeros. This prevents unsafe or misleading inputs like " " or "3e-2" from being treated as valid. Thanks for the heads‑up!

debugging/book-library/script.js
AngelaMcLeary: Hi @cjyuan, thank you for your feedback and suggestion. I have updated the code to prevent any HTML or script injection issues.

debugging/book-library/script.js
```diff
@@ -28,33 +28,66 @@ const check = document.getElementById("check");
...
-    title.value == null ||
-    title.value == "" ||
-    pages.value == null ||
-    pages.value == ""
+    !titleValue ||
+    !authorValue ||
+    !pagesValue ||
+    Number(pagesValue) <= 0 ||
+    Number.isNaN(Number(pagesValue)) ||
+    !/^[1-9]\d*$/.test(pagesValue)
```
cjyuan, line 40: Instead of repeatedly converting `pagesValue` to a number, wouldn't it be easier to save the converted number in a variable first?

When it is a number, you then check its range and check if it is an integer (using `Number.isinteger()`).

cjyuan (commented):
Changes look good.

illicitonion:

PR #433, opened 2026-04-13
Manchester | 26-ITP-Jan | Mehroz Munir | Sprint 2| Book Library

MehrozMunir:
## Learners, PR Template

Self checklist

## Changelist

I have tried to make the book library website more interactive, added validation for the input fields, and resolved the bugs in rendering and adding data.

### DIFF
--- a/debugging/book-library/index.html
+++ b/debugging/book-library/index.html
@@ -1,20 +1,20 @@
 <!DOCTYPE html>
-<html>
+<html lang="en">
   <head>
-    <title> </title>
     <meta
-      charset="utf-8"
+      charset="utf-8" >
+    <meta
       name="viewport"
-      content="width=device-width, initial-scale=1.0"
-    />
+      content="width=device-width, initial-scale=1.0" >
+    <title>My Book Library</title>
     <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.4.1/jquery.min.js"></script>
     <script src="https://cdnjs.cloudflare.com/ajax/libs/popper.js/1.16.0/umd/popper.min.js"></script>
     <script src="https://maxcdn.bootstrapcdn.com/bootstrap/4.4.1/js/bootstrap.min.js"></script>
     <link
       rel="stylesheet"
       href="https://maxcdn.bootstrapcdn.com/bootstrap/4.4.1/css/bootstrap.min.css"
-    />
-    <link rel="stylesheet" type="text/css" href="style.css" />
+    >
+    <link rel="stylesheet" type="text/css" href="style.css" >
   </head>
 
   <body>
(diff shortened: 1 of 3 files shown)


debugging/book-library/index.html
cjyuan, line 1: According to https://validator.w3.org/, there is an error in your `index.html`. Can you fix it?

debugging/book-library/script.js
```diff
@@ -1,14 +1,13 @@
 let myLibrary = [];
```
cjyuan: Could consider declaring it using `const`.

debugging/book-library/script.js
```diff
@@ -20,26 +19,38 @@ function populateStorage() {
...
   ) {
     alert("Please fill all fields!");
     return false;
   } else {
-    let book = new Book(title.value, title.value, pages.value, check.checked);
-    library.push(book);
+    const book = new Book(
+      titleInput.value,
+      authorInput.value,
+      pagesInput.value,
```
cjyuan, line 47: `.value` is a string, and a user can enter a value such as `"1.4e2"` (a numerical value equals to 140) in the input field.

debugging/book-library/script.js
```diff
@@ -51,51 +62,49 @@ function Book(title, author, pages, check) {
...
+    changeButton.className = "btn btn-success";
+    wasReadCell.appendChild(changeButton);
     let readStatus = "";
-    if (myLibrary[i].check == false) {
-      readStatus = "Yes";
-    } else {
-      readStatus = "No";
-    }
-    changeBut.innerText = readStatus;
+    myLibrary[i].check ? (readStatus = "Yes") : (readStatus = "No");
```
cjyuan: Normally way to use `? :` is:

`readStatus = myLibrary[i].check ? "Yes" : "No";`

`condition ? v1 : v2` is an **expression**. It evaluates to `v1` if `condition` is true, and to `v2` `if condition` is false.

debugging/book-library/script.js
MehrozMunir: Yes, that's true. I have resolved this issue by converting this back to a number when rendering this value, as shown here

` pagesCell.textContent = Number(myLibrary[i].pages); //line 81 of script.js`

MehrozMunir:
@cjyuan  Thank you for your useful feedback. I have made the changes you mentioned in the comments. Can you please review the code now.
Thanks

debugging/book-library/script.js
```diff
@@ -51,51 +62,49 @@ function Book(title, author, pages, check) {
...
-    pagesCell.innerHTML = myLibrary[i].pages;
+    const row = tableBody.insertRow();
+    const titleCell = row.insertCell(0);
+    const authorCell = row.insertCell(1);
+    const pagesCell = row.insertCell(2);
+    const wasReadCell = row.insertCell(3);
+    const deleteCell = row.insertCell(4);
+    titleCell.textContent = myLibrary[i].title;
+    authorCell.textContent = myLibrary[i].author;
+    pagesCell.textContent = Number(myLibrary[i].pages);
```
cjyuan: This may work for this app, but it is a dangerous practice.

A better place to ensure `pages` are number is in the Book constructor (line 60).

cjyuan (commented):
Changes look good.

debugging/book-library/script.js
MehrozMunir: Yes thanks for highlighting that. 
I have done that now.

`this.pages = Number(pages);`

illicitonion:

PR #437, opened 2026-04-14
Sheffield | 26-ITP-jan | Richard Frimpong | Sprint 2 | Book Library

Richiealx:
## Learners, PR Template

Self checklist

- [x] My changes follow the style guide

## Changelist

This PR fixes the bugs in the Book Library debugging project for Sprint 2 of the Data Flows module.

Changes made:
- fixed initial rendering so books display on page load
- fixed form submission so new books are added to `myLibrary`
- fixed validation to require title, author, and a valid page count
- fixed author field bug so the correct author is stored
- stored page count as a number instead of a string
- fixed read status so it displays correctly
- fixed toggle button behaviour for read/unread
- fixed delete functionality so books are removed correctly
- improved variable naming for DOM elements
- cleaned up table rendering to use `tbody`
- corrected invalid HTML input types and improved basic HTML validation

### DIFF
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -7,48 +7,47 @@ window.addEventListener("load", function () {
 
 function populateStorage() {
   if (myLibrary.length === 0) {
-    const book1 = new Book("Robinson Crusoe", "Daniel Defoe", "252", true);
+    const book1 = new Book("Robinson Crusoe", "Daniel Defoe", 252, true);
     const book2 = new Book(
       "The Old Man and the Sea",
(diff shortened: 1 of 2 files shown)


debugging/book-library/script.js
```diff
@@ -1,46 +1,55 @@
 let myLibrary = [];
```
cjyuan, line 1: Could declare this using `const`.

cjyuan (commented):
Looks good. Well done.

illicitonion:

PR #441, opened 2026-04-15
Manchester | ITP Jan-26 | Ofonime Edak | Sprint 2 | Book library

Ofonimeedak:
Self checklist

Debug and fix errors in the book library app

### DIFF
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -2,7 +2,7 @@ let myLibrary = [];
 
 window.addEventListener("load", function (e) {
   populateStorage();
-  render();
+  //render();
 });
 
 function populateStorage() {
@@ -32,16 +32,21 @@ function submit() {
     title.value == null ||
     title.value == "" ||
     pages.value == null ||
-    pages.value == ""
+    pages.value == "" ||
+    author.value == null ||
+    author.value == ""
   ) {
     alert("Please fill all fields!");
     return false;
   } else {
-    let book = new Book(title.value, title.value, pages.value, check.checked);
-    library.push(book);
+    let book = new Book(title.value, author.value, pages.value, check.checked);
+    myLibrary.push(book);
     render();
   }
 }
+
+const submitInput = document.getElementById("submit");
+submitInput.addEventListener("click", submit);
 
 function Book(title, author, pages, check) {
   this.title = title;
@@ -54,7 +59,7 @@ function render() {
   let table = document.getElementById("display");
   let rowsNumber = table.rows.length;
   //delete old table
-  for (let n = rowsNumber - 1; n > 0; n-- {
+  for (let n = rowsNumber - 1; n >= 1; n--) {
     table.deleteRow(n);
   }
   //insert updated row and cells
@@ -76,11 +81,12 @@ function render() {
     changeBut.className = "btn btn-success";
     wasReadCell.appendChild(changeBut);
     let readStatus = "";
-    if (myLibrary[i].check == false) {
+    if (myLibrary[i].check == true) {
       readStatus = "Yes";
     } else {
       readStatus = "No";
     }
+
     changeBut.innerText = readStatus;
 
     changeBut.addEventListener("click", function () {
@@ -89,12 +95,14 @@ function render() {
     });
 
     //add delete button to every row and render again
+
     let delButton = document.createElement("button");
-    delBut.id = i + 5;
-    deleteCell.appendChild(delBut);
-    delBut.className = "btn btn-warning";
-    delBut.innerHTML = "Delete";
-    delBut.addEventListener("clicks", function () {
+    delButton.id = i;
+
+    deleteCell.appendChild(delButton);
+    delButton.className = "btn btn-warning";
+    delButton.innerHTML = "Delete";
+    delButton.addEventListener("click", function () {
       alert(`You've deleted title: ${myLibrary[i].title}`);
       myLibrary.splice(i, 1);
       render();
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -20,26 +20,36 @@ function populateStorage() {
   }
 }
 
-const title = document.getElementById("title");
-const author = document.getElementById("author");
-const pages = document.getElementById("pages");
-const check = document.getElementById("check");
+const bookTitle = document.getElementById("title");
+const authorOfBook = document.getElementById("author");
+const NoOfpages = document.getElementById("pages");
+const readBook = document.getElementById("check");
 
 //check the right input from forms and if its ok -> add the new book (object in array)
 //via Book function and start render function
 function submit() {
   if (
-    title.value == null ||
-    title.value == "" ||
-    pages.value == null ||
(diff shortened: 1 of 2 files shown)


cjyuan (commented):

debugging/book-library/script.js
cjyuan: Safer to declare `myLibrary` using `const`.

debugging/book-library/script.js
```diff
@@ -2,7 +2,7 @@ let myLibrary = [];
 
 window.addEventListener("load", function (e) {
   populateStorage();
-  render();
+  //render();
```
cjyuan: If the code is not needed, remove it instead of just comment it out (to keep the code clean).

debugging/book-library/script.js
```diff
@@ -20,41 +20,56 @@ function populateStorage() {
...
 }
 
-const title = document.getElementById("title");
-const author = document.getElementById("author");
-const pages = document.getElementById("pages");
-const check = document.getElementById("check");
+const bookTitle = document.getElementById("title");
+const authorOfBook = document.getElementById("author");
+const NoOfpages = document.getElementById("pages");
+const readBook = document.getElementById("check");
```
cjyuan: - Can you give them names that can better reflect the values assigned to them -- objects representing HTML elements?

- Common practice is to declare all shared constants/variables at the beginning of the file (before any function definition).

- In JS convention, names that begin with an uppercase letter are used for data types. (e.g., Math, Book, Array).

debugging/book-library/script.js
```diff
@@ -20,41 +20,56 @@ function populateStorage() {
...
+    return false;
   } else {
-    let book = new Book(title.value, title.value, pages.value, check.checked);
-    library.push(book);
+    let book = new Book(
+      bookTitle.value,
+      authorOfBook.value,
+      NoOfpages.value,
+      readBook.checked
+    );
```
cjyuan, line 40: - You checked the trimmed values but you used the untrimmed values. As a result, the book's title and author's name can still contain leading and trailing space.

- Even if the HTML input type is "number", its value is still treated as a string, allowing inputs like "3e2" or "-1".

A better and safer approach to deal with user input is to first preprocess/sanitise them, and then store the cleaned values in variables. Thereafter, refer only to the variables for cleaned values consistently throughout the rest of the code.

debugging/book-library/script.js
```diff
@@ -20,41 +20,56 @@ function populateStorage() {
...
   }
 }
 
-function Book(title, author, pages, check) {
-  this.title = title;
-  this.author = author;
-  this.pages = pages;
-  this.check = check;
+const submitInput = document.getElementById("submit");
+submitInput.addEventListener("click", submit);
```
cjyuan: It's a better practice to keep all the code that is to be executed once on page load in the same place. For example, in the window "on load" event listener.

debugging/book-library/script.js
```diff
@@ -20,41 +20,56 @@ function populateStorage() {
...
 }
 
 function render() {
   let table = document.getElementById("display");
   let rowsNumber = table.rows.length;
   //delete old table
-  for (let n = rowsNumber - 1; n > 0; n-- {
+  for (let n = rowsNumber - 1; n >= 1; n--) {
     table.deleteRow(n);
   }
```
cjyuan, line 67: Clearing `<tbody>` is easier and more efficient.

debugging/book-library/script.js
```diff
@@ -20,41 +20,56 @@ function populateStorage() {
...
-  this.check = check;
+const submitInput = document.getElementById("submit");
+submitInput.addEventListener("click", submit);
+
+function Book(bookTitle, authorOfBook, NoOfpages, readBook) {
+  this.bookTitle = bookTitle;
+  this.authorOfBook = authorOfBook;
+  this.NoOfpages = NoOfpages;
+  this.readBook = readBook;
 }
```
cjyuan: - The resulting "Book" object already represents a book, so including the word "book" in its property names is redundant.

- Consider also using the conventional name for properties that store a Boolean value.

debugging/book-library/script.js
```diff
@@ -66,36 +81,34 @@ function render() {
     let pagesCell = row.insertCell(2);
     let wasReadCell = row.insertCell(3);
     let deleteCell = row.insertCell(4);
-    titleCell.innerHTML = myLibrary[i].title;
-    authorCell.innerHTML = myLibrary[i].author;
-    pagesCell.innerHTML = myLibrary[i].pages;
+    titleCell.innerHTML = myLibrary[i].bookTitle;
+    authorCell.innerHTML = myLibrary[i].authorOfBook;
+    pagesCell.innerHTML = myLibrary[i].NoOfpages;
```
cjyuan: Please note that when setting the text content of an HTML element, there are subtle but important differences between using `.innerHTML`, `innerText`, and `textContent`.

debugging/book-library/script.js
```diff
@@ -66,36 +81,34 @@ function render() {
...
-    wasReadCell.appendChild(changeBut);
-    let readStatus = "";
-    if (myLibrary[i].check == false) {
-      readStatus = "Yes";
-    } else {
-      readStatus = "No";
-    }
-    changeBut.innerText = readStatus;
+    let changeBtn = document.createElement("button");
+    changeBtn.id = i;
```
cjyuan: - Why give `changeBtn` an `id` attribute but not doing the same for `delButton`?
- Is it necessary to give them an `id` attribute in this app?

debugging/book-library/script.js
```diff
@@ -66,36 +81,34 @@ function render() {
...
-    changeBut.addEventListener("click", function () {
-      myLibrary[i].check = !myLibrary[i].check;
+    changeBtn.addEventListener("click", function () {
+      myLibrary[i].readBook = !myLibrary[i].readBook;
       render();
     });
 
     //add delete button to every row and render again
+
     let delButton = document.createElement("button");
```
cjyuan: Can you suggest a more consistent naming convention for the variables representing the two buttons, currently named `changeBut` and `delButton`?

cjyuan: Could consider declaring the two buttons using `const`.

debugging/book-library/script.js
Ofonimeedak: Thank you, I was thinking that breaking responsibility into each function will be easier to maintain, and could be the modern approach

debugging/book-library/script.js
Ofonimeedak: innerText is most appropriate because we want to keep the HTML style for that element and not override it using textContent

debugging/book-library/script.js
Ofonimeedak: not needed element already  created in the DOM memory, not gotten from the HTML page

debugging/book-library/script.js
cjyuan: Why not adopt it if it is the most appropriate one?

cjyuan:
Changes look good.

illicitonion:

PR #450, opened 2026-04-16
Sheffield | 26-ITP-Jan | Mahmoud Shaabo | Sprint 2 | Book Library

mahmoudshaabo1984:
## What bugs did I fix?

- [x] Bug 1 - Website loads but doesn't show any books
The `for` loop in the `render` function was missing a closing parenthesis `)` before the `{`. This caused a syntax error that stopped all JavaScript from running.

- [x] Bug 2 - Error in console when you try to add a book
The `submit` function was calling `library.push(book)` but the array is named `myLibrary`. Fixed by using the correct variable name. Also added validation for the author field.

- [x] Bug 3 - It uses the title name as the author name
In the `submit` function, `title.value` was used twice. The second argument should be `author.value`.

- [x] Bug 4 - Delete button is broken
The delete button variable was created as `delButton` but then referenced as `delBut` three times. Also the event name was `"clicks"` instead of `"click"`. Fixed both issues.

- [x] Bug 5 - Read status saves the wrong answer
The `if` condition was inverted. `check == false` was showing "Yes". Fixed so `check == true` correctly shows "Yes".

## How to test

1. Open `index.html` in the browser
2. Two books appear on load — both show "Yes" in the Read column
3. Click "Add new book" and fill in the form — the new book appears in the table
4. Click the "Yes"/"No" button on any book — it toggles correctly
5. Click "Delete" on any book — the book is removed from the table
6. Try submitting the form with empty fields — an alert appears

Hi @cifarquhar — I have fixed all 5 bugs listed in the readme. Please let me know if you have any feedback. Thank you!


mahmoudshaabo1984:
Hi @cifarquhar — PR #450 is ready for review.

I have fixed all 5 bugs listed in the readme.md file. The app now loads correctly, books can be added and deleted, and the read status toggles as expected.

Please let me know if anything needs to be changed. Thank you!

cjyuan (commented):
You have fixed all the bugs, but there are still plenty of room you can improve the code.

mahmoudshaabo1984:
Hi CJ,

Thank you for your feedback. I have reviewed your comments and applied the following improvements:

**script.js**
- Changed `let` to `const` for all DOM references that never change
- Renamed confusing variables: `check` → `hasBeenRead`, `title` → `titleInput`, `author` → `authorInput`
- Renamed `submit()` to `addBook()` to avoid conflict with the browser's built-in submit behaviour
- Replaced `innerHTML` with `textContent` for all user-provided data to prevent XSS vulnerabilities
- Renamed `populateStorage()` to `addDefaultBooks()` — the old name implied localStorage usage
- Extracted `createReadButton()` and `createDeleteButton()` as separate helper functions to make `render()` easier to read
- Added form field clearing after a book is successfully added

**index.html**
- Added `lang="en"` to the `<html>` element for accessibility
- Added a proper page title: "My Book Library"
- Fixed invalid input types: `type="title"` and `type="author"` changed to `type="text"`

Please let me know if there is anything else to improve.

Thank you,
Mahmoud Shaabo

debugging/book-library/script.js
```diff
@@ -1,103 +1,136 @@
+// myLibrary is the main array — like a shelf that holds all book objects
 let myLibrary = [];
```
cjyuan: Safer to declare `myLibrary` using `const` because it is not supposed to be reassigned a value.

debugging/book-library/script.js
```diff
@@ -1,103 +1,136 @@
...
 
-window.addEventListener("load", function (e) {
-  populateStorage();
+// Use const for DOM references because these elements never change
+// "const" means: this variable will always point to the same element
+const titleInput = document.getElementById("title");
+const authorInput = document.getElementById("author");
+const pagesInput = document.getElementById("pages");
+const readCheckbox = document.getElementById("check");
```
cjyuan, line 10: Good practice to declare all shared constants/variables at the beginning of the file.

debugging/book-library/script.js
```diff
@@ -1,103 +1,136 @@
...
-      "127",
-      true
+// Add two default books so the page is not empty on first load
+// Renamed from "populateStorage" — that name implied localStorage usage
+function addDefaultBooks() {
+  if (myLibrary.length === 0) {
+    myLibrary.push(new Book("Robinson Crusoe", "Daniel Defoe", "252", true));
+    myLibrary.push(
+      new Book("The Old Man and the Sea", "Ernest Hemingway", "127", true)
     );
```
cjyuan: It is better to use an appropriate data type to represent all data. 
Could consider storing "page count" as a value of type "number".

debugging/book-library/script.js
```diff
@@ -1,103 +1,136 @@
...
+function addBook() {
+  // Validate that all fields are filled before adding the book
   if (
-    title.value == null ||
-    title.value == "" ||
-    pages.value == null ||
-    pages.value == ""
+    titleInput.value === "" ||
+    authorInput.value === "" ||
+    pagesInput.value === ""
```
cjyuan: - String input may contain only space characters or leading/trailing space. Ideally the app should not accept a string containing only spaces as a book title or author's name.

- `pageInput.value` is a string and it can be a number in scientific format (e.g. "3e2") -- which may look weird when displayed as a value of page count.

A better and safer approach to deal with user input is to first preprocess/sanitise them, and then store the cleaned values in variables. Thereafter, refer only to the variables for cleaned values consistently throughout the rest of the code.

debugging/book-library/script.js
```diff
@@ -1,103 +1,136 @@
...
+function createDeleteButton(index) {
+  const button = document.createElement("button");
+  button.className = "btn btn-warning";
+  button.textContent = "Delete";
+
+  button.addEventListener("click", function () {
+    // splice(index, 1) removes exactly 1 item at position "index"
+    myLibrary.splice(index, 1);
+    render();
+  });
```
cjyuan, line 114: The original code has a statement that uses `window.alert()` to show the title of the deleted book whenever a book is deleted.

debugging/book-library/script.js
```diff
@@ -1,103 +1,136 @@
...
+  button.textContent = myLibrary[index].hasBeenRead ? "Yes" : "No";
+
+  button.addEventListener("click", function () {
+    // Flip the read status: true becomes false, false becomes true
+    myLibrary[index].hasBeenRead = !myLibrary[index].hasBeenRead;
+    render();
+  });
+
+  return button;
+}
```
cjyuan, line 101: I like the way you use function to isolate the logic for creating the the Read and Delete buttons.

debugging/book-library/script.js
```diff
@@ -1,103 +1,136 @@
...
-  //delete old table
-  for (let n = rowsNumber - 1; n > 0; n-- {
-    table.deleteRow(n);
+  const table = document.getElementById("display");
+  const rowCount = table.rows.length;
+
+  // Delete all rows except the header row (row 0)
+  for (let i = rowCount - 1; i > 0; i--) {
+    table.deleteRow(i);
   }
```
cjyuan: For better performance (and also simpler code), you can consider clearing `<tbody>` instead.

mahmoudshaabo1984:
Hi CJ,

Thank you for the detailed second round of feedback. I have applied all of your suggestions:

1. Changed `let myLibrary` to `const myLibrary` — it is never reassigned, only mutated.

2. Changed page count from a string ("252") to a number (252) — correct data type for numeric data.

3. Added input sanitisation before validation:
   - `sanitiseText()` trims leading and trailing whitespace, preventing inputs with only spaces from being accepted.
   - `sanitisePages()` rejects scientific notation like "3e2" and converts the value to a safe integer using `Math.floor()`.

4. Restored the `alert()` that shows the title of a deleted book when the Delete button is clicked.

5. Replaced the row-by-row deletion loop with `tbody.innerHTML = ""` — simpler and more performant.

Please let me know if there is anything else to improve.

Thank you,
Mahmoud Shaabo

debugging/book-library/script.js
```diff
@@ -1,103 +1,142 @@
...
+function createDeleteButton(index) {
+  const button = document.createElement("button");
+  button.className = "btn btn-warning";
+  button.textContent = "Delete";
+
+  button.addEventListener("click", function () {
+    // Show an alert with the deleted book's title — restoring original behaviour
+    alert(`"${myLibrary[index].title}" has been deleted.`);
+    myLibrary.splice(index, 1);
+    render();
```
cjyuan, line 113: The alert message is shown before the book is actually deleted; the deletion only occurs after the alert dialog is dismissed. This introduces a risk that the operation may not complete (e.g., if the user closes the browser before dismissing the alert).

debugging/book-library/script.js
```diff
@@ -1,103 +1,142 @@
...
+}
+
+// sanitisePages converts the pages input to a safe integer
+// Rejects scientific notation like "3e2" which would display strangely
+function sanitisePages(value) {
+  const trimmed = value.trim();
+  // If the input contains "e" or "E", it is scientific notation — reject it
+  if (trimmed.toLowerCase().includes("e")) {
+    return NaN;
   }
```
cjyuan, line 50: Can also just let `Number()` convert the input even if it contains an 'e'  -- less code to maintain.

cjyuan (commented):
Changes look good. Well done.

illicitonion:

PR #452, opened 2026-04-16
London | Jan26 | Ihor Taradaiko | Sprint 2 | Book Library

ihortar:
Self checklist

## Changelist

Fixed:
1. Website loads but doesn't show any books
2. Error in console when you try to add a book
3. It uses the title name as the author name
4. Delete button is broken
5. When I add a book that I say I've read - it saves the wrong answer


cjyuan (commented):
Please refer to the latest comment instead.

debugging/book-library/script.js
```diff
@@ -1,22 +1,22 @@
 let myLibrary = [];
```
cjyuan: Safer to declare this using `const`.

debugging/book-library/script.js
```diff
@@ -25,22 +25,24 @@ const author = document.getElementById("author");
 const pages = document.getElementById("pages");
 const check = document.getElementById("check");
```
cjyuan: - Common practice is to declare shared variables/constants at the beginning of the file (before function definition).

- Could also rename them to better reflect that the values assigned to them are HTML objects.

debugging/book-library/script.js
```diff
@@ -25,22 +25,24 @@ const author = document.getElementById("author");
...
     alert("Please fill all fields!");
-    return false;
-  } else {
-    let book = new Book(title.value, title.value, pages.value, check.checked);
-    library.push(book);
-    render();
+    return;
   }
+
+  let book = new Book(title.value, author.value, pages.value, check.checked);
```
cjyuan: - The app checks the trimmed value but uses the untrimmed values. 

- The title and author input could contain leading/trailing space.

- `page.value` could be a string containing value like "3e2" or "-1".

debugging/book-library/script.js
```diff
@@ -51,53 +53,46 @@ function Book(title, author, pages, check) {
...
-    delBut.innerHTML = "Delete";
-    delBut.addEventListener("clicks", function () {
+    delButton.className = "btn btn-warning";
+    delButton.innerHTML = "Delete";
+
+    delButton.addEventListener("click", function () {
       alert(`You've deleted title: ${myLibrary[i].title}`);
       myLibrary.splice(i, 1);
       render();
     });
```

cjyuan (commented):
You already have a good start.

debugging/book-library/script.js
ihortar: Edited to delete first, then show the message

debugging/book-library/script.js
ihortar: Moved to top but names seems ok to me

debugging/book-library/script.js
ihortar: Made it clean

debugging/book-library/script.js
```diff
@@ -1,46 +1,52 @@
...
-    let book = new Book(title.value, title.value, pages.value, check.checked);
-    library.push(book);
-    render();
+    return;
   }
+
+  let book = new Book(
+    cleanedTitle,
+    cleanedAuthor,
+    Number(cleanedPages),
```
cjyuan, line 40: When dealing with number input, we should consider its possible type and range.
Have you checked if `Number(cleanedPages)` is a positive integer ?

cjyuan (commented):
Changes look good. 

I will mark this as Complete first.

illicitonion:

PR #454, opened 2026-04-17
London | 26-ITP-Jan | Alex Okorefe | Sprint 2 | Book Library

Alex-Os-Dev-Lab:
…Book object

## Learners, PR Template

Self checklist

## Changelist

Completed tasks related to the debugging/book-library repo

## Questions

None at this stage

### DIFF
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -28,20 +28,38 @@ const check = document.getElementById("check");
 //check the right input from forms and if its ok -> add the new book (object in array)
 //via Book function and start render function
 function submit() {
-  if (
-    title.value === "" ||
-    author.value === "" ||
-    pages.value === "" ||
-    !check.checked //if the checkbox is not checked, it will be false, so we need to check if its not checked
-  ) {
-    alert("Please fill all fields!");
+  
+ // 1. Preprocess and store cleaned values in variables / removes spaces from the beginning and end
+  const cleanTitle = title.value.trim();
+  const cleanAuthor = author.value.trim();
+  
+  // Convert string input to a Number for calculation/validation
+  const pageCount = Number(pages.value); 
+  const isRead = check.checked;
+
+  // 2. VALIDATION: // Check if strings are empty after trimming spaces
+  if (cleanTitle === "" || cleanAuthor === "") {
+    alert("Title and Author cannot be empty!");
     return false;
-  } else {
-    let book = new Book(title.value, author.value, pages.value, check.checked);
-    myLibrary.push(book);
+  }
+  // Check if page count is a valid number and greater than 0
+  if (isNaN(pageCount) || pageCount <= 0) {
+    alert("Please enter a valid number of pages!");
+    return false;
+  }
+
+  // 3. If all validations pass, create a new Book object and add it to the library
+  let book = new Book(cleanTitle, cleanAuthor, pageCount, isRead);
+  myLibrary.push(book);
+
+  // 4. Clear the form fields after submission
+    title.value = "";
+    author.value = "";
+    pages.value = "";
+    check.checked = false;
+  
     render();
   }
-}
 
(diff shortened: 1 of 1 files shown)


cjyuan (commented):

debugging/book-library/script.js
```diff
@@ -1,6 +1,6 @@
 let myLibrary = [];
```
cjyuan, line 1: Can we declare `myLibrary` in a way that prevents it from being accidentally reassigned?

debugging/book-library/script.js
```diff
@@ -29,12 +29,10 @@ const check = document.getElementById("check");
...
   if (
-    title.value == null ||
-    title.value == "" ||
-    author.value == null ||
-    author.value == "" ||
-    pages.value == null ||
-    pages.value == ""
+    title.value === "" ||
+    author.value === "" ||
+    pages.value === "" ||
```
cjyuan: Should the app allow title/author to contains leading/trailing space, or even just space?

What type and range of numbers should a page count be?

debugging/book-library/script.js
cjyuan: These are raw input and can contain invalid values. For example, `pages.value` could be `"3e2"` (a string representation of a number in scientific form).

cjyuan (commented):

- Not all bugs have been fixed.

- There are quite some improvements you can still make, and I only highlighted some.
Please check this general feedback to find out how else you can further improve your code :

debugging/book-library/script.js
cjyuan: Can you address this comment?

debugging/book-library/script.js
```diff
@@ -66,17 +86,17 @@ function render() {
     let pagesCell = row.insertCell(2);
     let wasReadCell = row.insertCell(3);
     let deleteCell = row.insertCell(4);
```
cjyuan: These variables are also not going to be reassigned.

debugging/book-library/script.js
```diff
@@ -90,11 +110,11 @@ function render() {
 
     //add delete button to every row and render again
     let delButton = document.createElement("button");
-    delBut.id = i + 5;
-    deleteCell.appendChild(delBut);
-    delBut.className = "btn btn-warning";
-    delBut.innerHTML = "Delete";
-    delBut.addEventListener("clicks", function () {
+    delButton.id = i;
```
cjyuan: - Lines 95, 113:
  - `id` attribute should be unique. Are the values assigned to these `id` attributes unique? 

debugging/book-library/script.js
cjyuan: This could be a good opportunity to practice using the `? :` conditional operator. Can you rewrite the code on lines 98–104 as a single statement?

debugging/book-library/script.js
```diff
@@ -90,11 +110,11 @@ function render() {
...
+    deleteCell.appendChild(delButton);
+    delButton.className = "btn btn-warning";
+    delButton.innerHTML = "Delete";
+    delButton.addEventListener("click", function () {
       alert(`You've deleted title: ${myLibrary[i].title}`);
       myLibrary.splice(i, 1);
       render();
```

debugging/book-library/script.js
```diff
@@ -28,33 +28,53 @@ const check = document.getElementById("check");
...
 }
 
 function render() {
   let table = document.getElementById("display");
   let rowsNumber = table.rows.length;
   //delete old table
-  for (let n = rowsNumber - 1; n > 0; n-- {
+  for (let n = rowsNumber - 1; n > 0; n--) {
     table.deleteRow(n);
   }
```
cjyuan: It is easier and more efficient just to clear `<tbody>`; no need to use loop.

cjyuan (commented):

illicitonion:

PR #457, opened 2026-04-19
Sheffield | ITP-Jan-26 | Hayriye Saricicek | Sprint 2 | Book Library

mshayriyesaricicek:
Self checklist

## Changelist

I made changes to make the drop down form work and upload the book details to the book list.  I put an Enter button on the form.  If all books are deleted it will automatically load up the preloaded books.  If books are added it will save the information so it is available on reload.  All information is in the correct place.

### DIFF
--- a/debugging/book-library/index.html
+++ b/debugging/book-library/index.html
@@ -1,12 +1,9 @@
 <!doctype html>
-<html>
+<html lang="en">
   <head>
-    <title> </title>
-    <meta
-      charset="utf-8"
-      name="viewport"
-      content="width=device-width, initial-scale=1.0"
-    />
+    <title>Book Library</title>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.4.1/jquery.min.js"></script>
     <script src="https://cdnjs.cloudflare.com/ajax/libs/popper.js/1.16.0/umd/popper.min.js"></script>
     <script src="https://maxcdn.bootstrapcdn.com/bootstrap/4.4.1/js/bootstrap.min.js"></script>
@@ -31,9 +28,22 @@
       <form id="bookForm">
         <div class="form-group">
           <label for="title">Title:</label>
-          <input type="text" class="form-control" id="title" required />
+          <input
+            type="text"
+            title="Letters, numbers and basic punctuation only"
+            class="form-control"
+            id="title"
+            required
+          />
+
           <label for="author">Author: </label>
-          <input type="text" class="form-control" id="author" required />
+          <input
+            type="text"
+            class="form-control"
+            id="author"
+            title="Only letters, spaces, apostrophes and hyphens allowed"
+            required
+          />
           <label for="pages">Pages:</label>
           <input type="number" class="form-control" id="pages" required />
           <label class="form-check-label">
--- a/debugging/book-library/index.html
+++ b/debugging/book-library/index.html
@@ -1,4 +1,4 @@
-<!DOCTYPE html>
+<!doctype html>
 <html>
   <head>
     <title> </title>
@@ -28,46 +28,21 @@
     </button>
 
     <div id="demo" class="collapse">
-      <div class="form-group">
-        <label for="title">Title:</label>
-        <input
-          type="title"
-          class="form-control"
-          id="title"
-          name="title"
-          required
-        />
-        <label for="author">Author: </label>
-        <input
-          type="author"
-          class="form-control"
-          id="author"
-          name="author"
-          required
-        />
-        <label for="pages">Pages:</label>
-        <input
-          type="number"
-          class="form-control"
-          id="pages"
-          name="pages"
-          required
-        />
-        <label class="form-check-label">
-          <input
-            type="checkbox"
-            class="form-check-input"
-            id="check"
-            value=""
-          />Read
-        </label>
-        <input
-          type="submit"
-          value="Submit"
-          class="btn btn-primary"
-          onclick="submit();"
-        />
-      </div>
+      <form id="bookForm">
+        <div class="form-group">
+          <label for="title">Title:</label>
+          <input type="text" class="form-control" id="title" required />
+          <label for="author">Author: </label>
+          <input type="text" class="form-control" id="author" required />
+          <label for="pages">Pages:</label>
+          <input type="number" class="form-control" id="pages" required />
+          <label class="form-check-label">
+            <input type="checkbox" class="form-check-input" id="check" />Read
+          </label>
+
+          <button type="submit" class="btn btn-primary">Enter</button>
+        </div>
+      </form>
     </div>
 
     <table class="table" id="display">
@@ -80,15 +55,7 @@
           <th></th>
         </tr>
       </thead>
-      <tbody>
-        <tr>
-          <td></td>
-          <td></td>
-          <td></td>
-          <td></td>
-          <td></td>
-        </tr>
-      </tbody>
+      <tbody></tbody>
     </table>
 
     <script src="script.js"></script>
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -1,4 +1,4 @@
-let myLibrary = JSON.parse(localStorage.getItem("myLibrary")) || [];
+const myLibrary = [];
 
 function isValidAuthor(text) {
   for (let char of text) {
@@ -29,7 +29,6 @@ function isValidTitle(text) {
   }
   return true;
 }
-// 2. then your other functions (populateStorage, render, etc)
 
 function saveLibrary() {
   localStorage.setItem("myLibrary", JSON.stringify(myLibrary));
@@ -53,12 +52,14 @@ document.getElementById("bookForm").addEventListener("submit", function (e) {
     return;
   }
 
-  populateStorage(title, author, pages, read);
+  const success = populateStorage(title, author, pages, read);
+
+  if (!success) return; // ❗ STOP if invalid
 
   render();
+  saveLibrary();
 
   this.reset();
-
   $("#demo").collapse("hide");
 });
 
@@ -66,9 +67,14 @@ document.getElementById("bookForm").addEventListener("submit", function (e) {
 //via Book function and start render function
 
 function populateStorage(title, author, pages, check) {
+  if (!/^\d+$/.test(pages) || Number(pages) < 1 || Number(pages) > 5000) {
+    alert("Pages must be a whole number between 1 and 5000");
+    return false;
+  }
+
   let newBook = new Book(title, author, pages, check);
   myLibrary.push(newBook);
-  saveLibrary();
+  return true;
 }
 
 function Book(title, author, pages, check) {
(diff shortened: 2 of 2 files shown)


debugging/book-library/index.html

debugging/book-library/script.js
```diff
@@ -1,47 +1,58 @@
...
-    library.push(book);
-    render();
-  }
-}
+  function populateStorage(title, author, pages, check) {
+    let newBook = new Book(title, author, pages, check);
+
+  myLibrary.push(newBook);
+  saveLibrary(); 
+  };
```
cjyuan: Why are there two definitions of `populateStorage()`? Lines 40-45, and lines 49-54.

The original `populateStorage()` is meant to create two dummy books initially (so that the person testing the app don't have to manually enter two books every time they reload the page). Perhaps instead of changing the behavior of the original `populuateStorage()`, you can introduce a `createAndSaveBook()` function that performs what your `populateStorage()` is doing now? This way, I would always have at least two books created when I test the app in an Incognito browser window.

debugging/book-library/script.js
```diff
@@ -1,47 +1,58 @@
...
+function addBookToTable(title, author, pages, read) {
+  const table = document.getElementById("display").getElementsByTagName("tbody")[0];
+
+  const row = table.insertRow();
+
+  row.insertCell(0).textContent = title;
+  row.insertCell(1).textContent = author;
+  row.insertCell(2).textContent = pages;
+  row.insertCell(3).textContent = read ? "Yes" : "No";
 }
```
cjyuan: This function is not being called anywhere in this script.

debugging/book-library/script.js
```diff
@@ -51,16 +62,20 @@ function Book(title, author, pages, check) {
...
+    table.innerHTML = "";
+
+  //let rowsNumber = table.rows.length;
   //delete old table
-  for (let n = rowsNumber - 1; n > 0; n-- {
-    table.deleteRow(n);
-  }
+  //for (let n = rowsNumber - 1; n > 0; n--) {
+  //  table.deleteRow(n);
+  //}
```
cjyuan: Please delete all debugging or unused code to keep the code clean.

debugging/book-library/script.js
```diff
@@ -1,47 +1,58 @@
...
-      "127",
-      true
-    );
-    myLibrary.push(book1);
-    myLibrary.push(book2);
-    render();
+ if (!title || !author || !pages) {
+    alert("Please fill all fields!");
+    return;
   }
```
cjyuan: Note:
- Both `title` and `author` could be a string containing only space or leading/trailing space.
- `pages`, as a string, could hold input values like "3e2" and "-1".

debugging/book-library/script.js
```diff
@@ -72,32 +87,53 @@ function render() {
...
       render();
     });
+ }
+} 
+ window.addEventListener("load", function () {
+  if (myLibrary.length === 0) {
+    populateStorage("Robison Crusoe", "Daniel Defoe", "252", true);
+    populateStorage( "The Old Man and the Sea", "Ernest Hemingway", "127", false);
+    saveLibrary();
   }
```
cjyuan: These are placeholder data designed to make testing the app easier.
Keeping them in a function would make excluding/including them easier during development.

cjyuan (commented):
Code is not properly indented, making it hard to read. Can you fix the Indention?

debugging/book-library/index.html
mshayriyesaricicek: I amended what I could on HTML.  validtor is complaining about trailing / but I take them off  but when I save the file they are put back on.

debugging/book-library/script.js
mshayriyesaricicek: That was a mistake.  I was moving things around copying and pasting and mistakenly ended up with it twice.  I have deleted one now.

debugging/book-library/script.js
mshayriyesaricicek: I have done, thank y ou.

debugging/book-library/script.js
mshayriyesaricicek: I have restricted what can be entered.  I initially restricted to alpha letters only but then after thinking I allowed hypens, spaces and aposgtrophes in names because sometimes they are needed ie in double barrel surnames or names such as O'Connor.  I also allowed input of punctuation and numbers for titles as they can be used in book titles and not allowing them would prevent certain book titles from being added.

debugging/book-library/index.html
cjyuan: Don't worry about the trailing `/`. Just focus on eliminating "Error" and "Warning". Those messages tagged "Info" are not errors.

mshayriyesaricicek: there are now no errors.

debugging/book-library/script.js
mshayriyesaricicek: it should have been deleted.  It is now.

debugging/book-library/script.js
mshayriyesaricicek: Ive made a function loadDefaultBooks.  Thank you.

debugging/book-library/script.js
```diff
@@ -1,46 +1,74 @@
-let myLibrary = [];
+let myLibrary = JSON.parse(localStorage.getItem("myLibrary")) || [];
```
cjyuan: - Safer to declare `myLibrary` using `const`.

- Better to perform loading data from `localStorage` inside window's onload callback or in some function where you can use try/catch block to handle errors.

debugging/book-library/script.js
```diff
@@ -1,46 +1,74 @@
...
 
-window.addEventListener("load", function (e) {
-  populateStorage();
-  render();
-});
+function isValidAuthor(text) {
+  for (let char of text) {
+    const isLetter =
+      (char >= "a" && char <= "z") || (char >= "A" && char <= "Z");
```
cjyuan, line 6: When you have time, consider exploring the subject "JavaScript regular expression (regex)". With regex, we can replace all the code in this function by one expression.

debugging/book-library/script.js
```diff
@@ -1,46 +1,74 @@
...
+    alert("Invalid author name");
+    return;
+  }
+
+  if (!isValidTitle(title)) {
+    alert("Invalid book title");
+    return;
+  }
+
+  populateStorage(title, author, pages, read);
```
cjyuan: - Currently, `pages` is assigned a string, and it can possibly be assigned an input string in the form "3e2". You can consider converting `pages` to an equivalent number in `populateStorage()` or in `Book()`.

- In addition, what range and type of numbers should be considered valid for a page count?

debugging/book-library/script.js
```diff
@@ -1,46 +1,74 @@
...
+
+  render();
+
+  this.reset();
```
cjyuan, line 62: This seems quite advanced. Are you familiar with the meaning of `this` in JS? 

In Piscine, you may be asked something like this:
- If you were not allowed to use `this`, how would you rewrite `this.reset()`?

debugging/book-library/script.js
cjyuan: Please note that when setting the text content of an HTML element, there are subtle but important differences between using `.innerHTML`, `innerText`, and `textContent`.

debugging/book-library/script.js
```diff
@@ -51,16 +79,13 @@ function Book(title, author, pages, check) {
 }
 
 function render() {
-  let table = document.getElementById("display");
-  let rowsNumber = table.rows.length;
-  //delete old table
-  for (let n = rowsNumber - 1; n > 0; n-- {
-    table.deleteRow(n);
-  }
+  const table = document.querySelector("#display tbody");
```
cjyuan: Could consider changing the variable name to better reflect that it is assigned a "tbody".

debugging/book-library/script.js
```diff
@@ -72,32 +97,51 @@ function render() {
...
-      alert(`You've deleted title: ${myLibrary[i].title}`);
+
+    deleteCell.appendChild(delBut);
+
+    delBut.addEventListener("click", function () {
       myLibrary.splice(i, 1);
+
+      saveLibrary();
+
       render();
```
cjyuan, line 130: The original code has a statement that uses `window.alert()` to show the title of the deleted book.

debugging/book-library/script.js
mshayriyesaricicek: thanks, I have changed to const and moved local storage

debugging/book-library/script.js
mshayriyesaricicek: I wasn't sure if we could use regex.  Thanks will look into it.

debugging/book-library/script.js
mshayriyesaricicek: I have changed it to only accept numbers between 1 and 5000.

debugging/book-library/script.js
mshayriyesaricicek: If I couldn’t use this.reset(), I would manually reset each form field by setting their values back to empty. For example:  document.getElementById("title").value = "";
document.getElementById("author").value = "";
document.getElementById("pages").value = "";
document.getElementById("check").checked = false;

debugging/book-library/script.js
mshayriyesaricicek: I have learned more about it now and found that innerText is safer for input but innerHTML is okay if not input so have changed these to innerText.

debugging/book-library/script.js
mshayriyesaricicek: Thank you,  I have changed it to tableBody

debugging/book-library/script.js
mshayriyesaricicek: I have put it back in the file.  Thank you for bringing this to my attention.

debugging/book-library/script.js
```diff
@@ -1,46 +1,80 @@
...
+  if (!/^\d+$/.test(pages) || Number(pages) < 1 || Number(pages) > 5000) {
+    alert("Pages must be a whole number between 1 and 5000");
     return false;
-  } else {
-    let book = new Book(title.value, title.value, pages.value, check.checked);
-    library.push(book);
-    render();
   }
+
+  let newBook = new Book(title, author, pages, check);
```
cjyuan, line 75: Why do you choose to keep the pages as string? It is not wrong, just unusual.

debugging/book-library/script.js
```diff
@@ -51,53 +85,74 @@ function Book(title, author, pages, check) {
...
+
+    deleteCell.appendChild(delBut);
+
+    delBut.addEventListener("click", function () {
       alert(`You've deleted title: ${myLibrary[i].title}`);
       myLibrary.splice(i, 1);
+
+      saveLibrary();
+
       render();
```

It’s better to display a confirmation message only after the associated operation has successfully completed.

Where should you place the `alert()` function call?

debugging/book-library/script.js
cjyuan: You can use any code as long as can explain the code when asked. In Piscine's interview, the interviewer may randomly pick a piece of code from your project and ask you to explain it.

debugging/book-library/script.js
mshayriyesaricicek: When I load it up on live server and delete a book the delete message comes up after deleting the book not before but I have moved it to a more suitable position.

cjyuan: `alert()` blocks execution of subsequent statements. To simulate the behavior, you can try this:

```
      throw new Error("simulate something wrong happening ...");
      myLibrary.splice(i, 1);

      saveLibrary();

      render();
```

The error will only throw after you close the alert dialog box.

debugging/book-library/script.js
mshayriyesaricicek: ive changed it now

cjyuan:
Changes look good.

illicitonion:

PR #458, opened 2026-04-19
Birmingham | ITP-Jan-26 | Ayodeji Ayorinde |  Sprint 2 | Book Library

Ayogit1:
## Learners, PR Template

Self checklist

## Changelist

Completion of My Book Library

### DIFF
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -12,9 +12,10 @@ const tableBody = document.getElementById("tableBody");
 window.addEventListener("load", () => {
   populateStorage();
   render();
+  submitBtn.addEventListener("click", submitBook);
 });
 
-submitBtn.addEventListener("click", submitBook);
+
 
 function Book(title, author, pages, check) {
   this.title = title;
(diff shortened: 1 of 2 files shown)


debugging/book-library/script.js
```diff
@@ -1,3 +1,121 @@
...
+const submitBtn = document.getElementById("submitBtn");
+const tableBody = document.getElementById("tableBody");
+
+// Initialize app
+window.addEventListener("load", () => {
+  populateStorage();
+  render();
+});
+
+submitBtn.addEventListener("click", submitBook);
```
cjyuan: Line 17 is also a "run once on page load" code. As such, it is better to keep in inside the window's onload callback.

debugging/book-library/script.js
```diff
@@ -1,3 +1,121 @@
...
+  // Clear the container in one operation
+  tableBody.innerHTML = "";
+
+  myLibrary.forEach((book, index) => {
+    const row = document.createElement("tr");
+
+    // Title Cell
+    const titleCell = document.createElement("td");
+    titleCell.textContent = book.title;
+    row.appendChild(titleCell);
```
cjyuan: Why not use the original approach to create a table cell? That is,
```javascript
    // Less code to write
```
instead of 

```javascript
    const titleCell = document.createElement("td");
    row.appendChild(titleCell);
```

debugging/book-library/script.js
cjyuan: Any unused code should be deleted to keep the code clean. If someone needs to original code, they can retrieve the code from a previous commit.

debugging/book-library/script.js
Ayogit1: Line 17 now inside the window's onload callback.

debugging/book-library/script.js
Ayogit1: Original table creation approach now used

debugging/book-library/script.js
Ayogit1: Unused code removed

cjyuan:
Changes look good.

illicitonion:

PR #473, opened 2026-04-22
Cape Town | 2026-ITP-Jan | Pretty Taruvinga | Sprint 2 | Book Library

Pretty548:
## Learners, PR Template

Self checklist

## Changelist

1. Fixed index.html
2. Fixed script.js

## Questions

None

### DIFF
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -1,4 +1,4 @@
-let myLibrary = [];
+const myLibrary = [];
 
 window.addEventListener("load", function () {
   populateStorage();
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -1,4 +1,8 @@
 const myLibrary = [];
+const titleInput = document.getElementById("title");
+const authorInput = document.getElementById("author");
+const pagesInput = document.getElementById("pages");
+const readCheckbox = document.getElementById("check");
 
 window.addEventListener("load", function () {
   populateStorage();
@@ -20,11 +24,6 @@ function populateStorage() {
     // removed extra render() here
   }
 }
-
-const titleInput = document.getElementById("title");
-const authorInput = document.getElementById("author");
-const pagesInput = document.getElementById("pages");
-const readCheckbox = document.getElementById("check");
 
 function submit() {
   const title = titleInput.value.trim();
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -1,46 +1,45 @@
 let myLibrary = [];
 
-window.addEventListener("load", function (e) {
+window.addEventListener("load", function () {
   populateStorage();
-  render();
+  render(); // only called once now
 });
 
 function populateStorage() {
-  if (myLibrary.length == 0) {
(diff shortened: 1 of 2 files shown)


debugging/book-library/script.js
```diff
@@ -1,46 +1,45 @@
 let myLibrary = [];
```

debugging/book-library/script.js
```diff
@@ -1,46 +1,45 @@
...
 }
 
-const title = document.getElementById("title");
-const author = document.getElementById("author");
-const pages = document.getElementById("pages");
-const check = document.getElementById("check");
+const titleInput = document.getElementById("title");
+const authorInput = document.getElementById("author");
+const pagesInput = document.getElementById("pages");
+const readCheckbox = document.getElementById("check");
```
cjyuan, line 6: Normal practice is to declare all shared variables/constants at the beginning of the file (before function definition).

debugging/book-library/script.js
```diff
@@ -51,53 +50,50 @@ function Book(title, author, pages, check) {
...
-  let rowsNumber = table.rows.length;
-  //delete old table
-  for (let n = rowsNumber - 1; n > 0; n-- {
-    table.deleteRow(n);
+  const table = document.getElementById("display");
+
+  // remove old rows (keep header row)
+  while (table.rows.length > 1) {
+    table.deleteRow(1);
   }
```
cjyuan: Note: Easier and more efficient to just clear `<tbody>`.

cjyuan (commented):
Code looks good. I will mark this as "Complete" first.

debugging/book-library/script.js
Pretty548: Good point! I’ve updated myLibrary to use const to prevent reassignment.

debugging/book-library/script.js
Pretty548: Thanks for pointing that out , I’ll move these declarations to the top of the file for better organization.

debugging/book-library/script.js
Pretty548: Thanks for the suggestion — I’ve updated the render function to clear and rebuild the <tbody> instead.

debugging/book-library/script.js
```diff
@@ -49,21 +49,17 @@ function Book(title, author, pages, check) {
...
-    const titleCell = row.insertCell(0);
-    const authorCell = row.insertCell(1);
-    const pagesCell = row.insertCell(2);
-    const wasReadCell = row.insertCell(3);
-    const deleteCell = row.insertCell(4);
+    const titleCell = document.createElement("td");
+    const authorCell = document.createElement("td");
+    const pagesCell = document.createElement("td");
+    const wasReadCell = document.createElement("td");
+    const deleteCell = document.createElement("td");
```
cjyuan: Why not use the original approach? That is,

```javascript
    let authorCell = row.insertCell(1);
```

Pretty548: Thanks! I’ve applied this approach — I’m now using row.insertCell() for each column (title, author, pages, read status, delete), which simplifies the DOM manipulation and keeps the structure clear.

illicitonion:

PR #482, opened 2026-04-23
London | 26-ITP-Jan | Boualem Larbi Djebbour | Sprint 2 | Book Library

djebsoft:
## Learners, PR Template

Self checklist

## Changelist

fixed the bugs 

## Questions

### DIFF
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -7,7 +7,7 @@ window.addEventListener("load", function (e) {
 
 function populateStorage() {
   if (myLibrary.length == 0) {
-    let book1 = new Book("Robison Crusoe", "Daniel Defoe", "252", true);
+    let book1 = new Book("Robinson Crusoe", "Daniel Defoe", "252", true);
     let book2 = new Book(
       "The Old Man and the Sea",
       "Ernest Hemingway",
@@ -16,7 +16,6 @@ function populateStorage() {
     );
     myLibrary.push(book1);
     myLibrary.push(book2);
-    render();
   }
 }
 
@@ -31,14 +30,16 @@ function submit() {
   if (
     title.value == null ||
     title.value == "" ||
+    author.value == null ||
+    author.value == "" ||
     pages.value == null ||
     pages.value == ""
   ) {
     alert("Please fill all fields!");
     return false;
   } else {
-    let book = new Book(title.value, title.value, pages.value, check.checked);
-    library.push(book);
+    let book = new Book(title.value, author.value, pages.value, check.checked);
+    myLibrary.push(book);
     render();
   }
 }
@@ -54,7 +55,7 @@ function render() {
   let table = document.getElementById("display");
   let rowsNumber = table.rows.length;
   //delete old table
-  for (let n = rowsNumber - 1; n > 0; n-- {
+  for (let n = rowsNumber - 1; n > 0; n--) {
     table.deleteRow(n);
   }
   //insert updated row and cells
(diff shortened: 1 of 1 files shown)


debugging/book-library/script.js
```diff
@@ -75,26 +76,21 @@ function render() {
...
     wasReadCell.appendChild(changeBut);
-    let readStatus = "";
-    if (myLibrary[i].check == false) {
-      readStatus = "Yes";
-    } else {
-      readStatus = "No";
-    }
-    changeBut.innerText = readStatus;
+
+    changeBut.innerText = myLibrary[i].check ? "Yes" : "No";
```
Luro91, line 80: Nice use of the ternary operator

debugging/book-library/script.js
```diff
@@ -31,14 +30,16 @@ function submit() {
...
     alert("Please fill all fields!");
     return false;
   } else {
-    let book = new Book(title.value, title.value, pages.value, check.checked);
-    library.push(book);
+    let book = new Book(title.value, author.value, pages.value, check.checked);
+    myLibrary.push(book);
     render();
   }
 }
```
Luro91, line 45: Before using input in an app, we should properly preprocess it (even if basic validation has already been enforced in HTML):
  - **Input Validation** - Check whether the input meets expected formats, types, and constraints
  - **Input Sanitization** - Clean input to exclude unwanted or potentially harmful data
  - **Input Normalization/Conversion** - Transform input into the correct data type or a standardized, usable format

Questions to consider:
  - Should `title` and `author` be allowed to contain only space characters?
  - Should `title` and `author` be allowed to contain leading or trailing space characters?
  - What type of value should we use to store the page count?
  - What kinds of input values (if any) should be rejected?

debugging/book-library/script.js
```diff
@@ -54,7 +55,7 @@ function render() {
   let table = document.getElementById("display");
   let rowsNumber = table.rows.length;
   //delete old table
-  for (let n = rowsNumber - 1; n > 0; n-- {
+  for (let n = rowsNumber - 1; n > 0; n--) {
     table.deleteRow(n);
   }
```
Luro91, line 60: ### Efficient approach to remove rows in `<table>`
In `render()`, instead of deleting table rows one by one, we could delete all rows in one operation by clearing their container.

debugging/book-library/script.js
```diff
@@ -75,26 +76,21 @@ function render() {
...
+    changeBut.innerText = myLibrary[i].check ? "Yes" : "No";
 
     changeBut.addEventListener("click", function () {
       myLibrary[i].check = !myLibrary[i].check;
       render();
     });
 
     //add delete button to every row and render again
-    let delButton = document.createElement("button");
+    let delBut = document.createElement("button");
```
Luro91, line 88: - Are the values assigned to the `id` attributes of the delete button unique? 
- Is there any need to assign an `id` attribute to the buttons?

debugging/book-library/script.js
```diff
@@ -54,7 +55,7 @@ function render() {
   let table = document.getElementById("display");
```
Luro91, line 55: When setting the text content of an HTML element, there are subtle but important differences between using `.innerHTML`, `innerText`, and `textContent`.

In `render()`, are the text values assigned to the appropriate property of each DOM node?

debugging/book-library/script.js
```diff
@@ -75,26 +76,21 @@ function render() {
...
-    let delButton = document.createElement("button");
+    let delBut = document.createElement("button");
     delBut.id = i + 5;
     deleteCell.appendChild(delBut);
     delBut.className = "btn btn-warning";
     delBut.innerHTML = "Delete";
-    delBut.addEventListener("clicks", function () {
+    delBut.addEventListener("click", function () {
       alert(`You've deleted title: ${myLibrary[i].title}`);
       myLibrary.splice(i, 1);
```
Luro91, line 95: In the callback function of the delete button, the alert message is shown before the book is actually deleted;
the deletion only occurs after the alert dialog is dismissed. This introduces a risk that the operation may
not complete (e.g., if the user closes the browser before dismissing the alert).

In general, it's better to display a confirmation message only after the associated operation has successfully
completed.

Luro91 (commented):
Good job on fixing the Major bugs.
There are still some improvements that can be done

According to https://validator.w3.org/, there are errors in your index.html. Can you fix these errors?

illicitonion:

PR #492, opened 2026-05-08
London | 26-ITP-May | Raihan Sharif | Sprint 1 | Destructuring exercises

RaihanSharif:
## Learners, PR Template

Self checklist

## Changelist
Did all required tasks. In exercise 2, ignore the last function if you just want a minimal working solution.

### DIFF
--- a/Sprint-1/destructuring/exercise-1/exercise.js
+++ b/Sprint-1/destructuring/exercise-1/exercise.js
@@ -6,7 +6,7 @@ const personOne = {
 
 // Update the parameter to this function to make it work.
 // Don't change anything else.
-function introduceYourself(___________________________) {
+function introduceYourself({ name, age, favouriteFood }) {
   console.log(
     `Hello, my name is ${name}. I am ${age} years old and my favourite food is ${favouriteFood}.`
   );
--- a/Sprint-1/destructuring/exercise-1/exercise.js
+++ b/Sprint-1/destructuring/exercise-1/exercise.js
@@ -6,7 +6,11 @@ const personOne = {
 
 // Update the parameter to this function to make it work.
 // Don't change anything else.
-function introduceYourself({ name, age, favouriteFood }) {
+function introduceYourself({
+  name = "NO NAME",
+  age = "NO AGE",
+  favouriteFood = "NO FOOD",
+}) {
   console.log(
     `Hello, my name is ${name}. I am ${age} years old and my favourite food is ${favouriteFood}.`
   );
--- a/Sprint-1/destructuring/exercise-1/exercise.js
+++ b/Sprint-1/destructuring/exercise-1/exercise.js
@@ -6,11 +6,7 @@ const personOne = {
 
 // Update the parameter to this function to make it work.
 // Don't change anything else.
-function introduceYourself({
-  name = "NO NAME",
-  age = "NO AGE",
-  favouriteFood = "NO FOOD",
-}) {
+function introduceYourself({ name, age, favouriteFood }) {
   console.log(
     `Hello, my name is ${name}. I am ${age} years old and my favourite food is ${favouriteFood}.`
   );
--- a/Sprint-1/destructuring/exercise-2/exercise.js
+++ b/Sprint-1/destructuring/exercise-2/exercise.js
@@ -100,9 +100,15 @@ function getPersonByPredicate(list, predicate) {
   });
 }
 
+/* 
+Example usable of the above general purpose version which allows the user to filter by different
+properties. 
+
 getPersonByPredicate(hogwarts, (person) => person.house === "Gryffindor");
 
 getPersonByPredicate(
   hogwarts,
   (person) => person.occupation === "Teacher" && person.pet
 );
+
+*/
--- a/Sprint-1/destructuring/exercise-2/exercise.js
+++ b/Sprint-1/destructuring/exercise-2/exercise.js
@@ -72,18 +72,16 @@ let hogwarts = [
 ];
 
 function logPeopleByHouse(peopleList, houseName) {
-  peopleList.forEach((person) => {
-    if (person.house === houseName) {
-      const { firstName, lastName } = person;
+  peopleList.forEach(({ house, firstName, lastName }) => {
+    if (house === houseName) {
       console.log(`${firstName} ${lastName}`);
     }
   });
 }
 
 function logTeachersWithPets(peopleList) {
-  peopleList.forEach((person) => {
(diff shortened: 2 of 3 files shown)


Sprint-1/destructuring/exercise-1/exercise.js
```diff
@@ -6,7 +6,11 @@ const personOne = {
 
 // Update the parameter to this function to make it work.
 // Don't change anything else.
-function introduceYourself(___________________________) {
+function introduceYourself({
+  name = "NO NAME",
```
illicitonion: Why did you decide to add default values here?

Sprint-1/destructuring/exercise-2/exercise.js
```diff
@@ -70,3 +70,45 @@ let hogwarts = [
     occupation: "Teacher",
   },
 ];
+
+function logPeopleByHouse(peopleList, houseName) {
```
illicitonion, line 74: These solutions work, but don't use destructuring as much as they could. Given this exercises is about destructuring, can you aim for maximum destructring?

Sprint-1/destructuring/exercise-3/exercise.js
```diff
@@ -6,3 +6,18 @@ let order = [
...
   { itemName: "Hash Brown", quantity: 4, unitPricePence: 40 },
 ];
+
+function printReceipt(order) {
+  console.log(`${"QTY".padEnd(8)}${"ITEM".padEnd(20)}TOTAL`);
+  let total = 0;
+  order.forEach((item) => {
+    const { quantity, itemName, unitPricePence } = item;
+    console.log(
+      `${String(quantity).padEnd(8)}${String(itemName).padEnd(20)}${((quantity * unitPricePence) / 100).toFixed(2)}`
```
illicitonion: There's quite a lot of repetition with your formatting. Imagine if we start to start padding QTY by only 4 or 12 characters - you'd need to update two different places. Can you work out how to reduce that repetition?

Sprint-1/destructuring/exercise-3/exercise.js
```diff
@@ -6,3 +6,18 @@ let order = [
   { itemName: "Hot Coffee", quantity: 2, unitPricePence: 100 },
   { itemName: "Hash Brown", quantity: 4, unitPricePence: 40 },
 ];
+
+function printReceipt(order) {
+  console.log(`${"QTY".padEnd(8)}${"ITEM".padEnd(20)}TOTAL`);
+  let total = 0;
+  order.forEach((item) => {
+    const { quantity, itemName, unitPricePence } = item;
```
illicitonion: It may be tidier to destructure in the parameter, rather than having a named parameter and them immediately destructuring it. WDYT?

illicitonion (commented):
This all works, but a few questions/suggestions about some details :)

Sprint-1/destructuring/exercise-1/exercise.js
RaihanSharif: I wanted users to be get some feedback if some values were not supplied, rather than just leaving it blank.

I have removed it.

Sprint-1/destructuring/exercise-2/exercise.js
RaihanSharif: I have added more destructuring, can this be considered maximum descructuring?

I prefer the previous solution in terms of readability. Though I guess if you have a lot of properties it can get tedious.

What do you think?

Sprint-1/destructuring/exercise-3/exercise.js
RaihanSharif: I agree, just doing the destructuring in the parameter makes it easier to read and understand. Code is now  changed.

Sprint-1/destructuring/exercise-3/exercise.js
RaihanSharif: I thought of adding a reciept formatter or string building function that would take quantity, itemName, and unitPricePence, and return a formatted String for console to output. I thought this might be overkill for a simple receipt function. 

Instead I just store the padding values up top as variables, so that padding can be easily adjusted, without having to touch the console log code.

Sprint-1/destructuring/exercise-2/exercise.js
illicitonion: Very reasonable - both what you've done, and your thoughts on the trade-offs of readability :)

Sprint-1/destructuring/exercise-3/exercise.js
illicitonion: Personally I'd lean towards a simple formatter function as you describe - it reduces e.g. the repetition of calling `padEnd` multiple times :) Want to give it a try?

Sprint-1/destructuring/exercise-3/exercise.js
```diff
@@ -7,18 +7,28 @@ let order = [
   { itemName: "Hash Brown", quantity: 4, unitPricePence: 40 },
 ];
 
-function printReceipt(order) {
+// helper function that pads each item in a line by a certain amount
+function getReceiptString(quantity, itemName, total) {
```
illicitonion: I'd maybe call this `formatReceiptEntry`

`get` suggests it's getting something that already exists, rather than creating it

`String` is very general, and my first read of `getReceiptString` suggests it's getting a string of the whole receipt - `entry` (or `line`) makes clear it's just one entry.

illicitonion (commented):
Looks good! One naming thought

Sprint-1/destructuring/exercise-3/exercise.js
RaihanSharif: I've moved string building to a helper function. 

I've tried to be very deliberate about my choices, considering alternatives. I thought about making `qtyWidth` and `itemWidth` into parameters of the string builder function, because I didn't want to hard code the widths. In the end, I felt it serves a different purpose to the supplying the quantity, name and total values, and so decided against it. 

I also tried to make sure the string building didn't do any 'business logic", just formatting. In the end I was able to make it general enough to print both the header and the line items. 

I also didn't use any extra functions for the total, as I felt that was simple enough.

Sprint-1/destructuring/exercise-3/exercise.js
RaihanSharif: Yes, I was thinking exactly the same thing, you got in there before I could edit :)

Naming things is one of the things I'm still trying to get to grips with. I was going to name it something like `buildLineItemString` or `formatLineItem`. Yours makes more sense.

PR #497, opened 2026-05-14
London | 26-ITP-May | Raihan Sharif | Sprint 2 | Book Library

RaihanSharif:
## Learners, PR Template

Self checklist

## Changelist
Books can be added or deleted
Books can be marked as completed or not at the time of adding them to list
Books can be marked as completed or not after adding

Code is checked and updated for quality against the feedback guideline here: https://github.com/CodeYourFuture/Module-Data-Flows/blob/general-review-feedback/debugging/book-library/feedback.md

### DIFF
--- a/debugging/book-library/index.html
+++ b/debugging/book-library/index.html
@@ -1,4 +1,4 @@
-<!DOCTYPE html>
+<!doctype html>
 <html>
   <head>
     <title> </title>
@@ -59,13 +59,13 @@
             class="form-check-input"
             id="check"
             value=""
-          />Read
+          />completed
         </label>
         <input
           type="submit"
           value="Submit"
           class="btn btn-primary"
-          onclick="submit();"
+          onclick="submit()"
         />
       </div>
     </div>
@@ -76,7 +76,7 @@
           <th>Title</th>
           <th>Author</th>
           <th>Number of Pages</th>
-          <th>Read</th>
+          <th>Completed</th>
           <th></th>
         </tr>
       </thead>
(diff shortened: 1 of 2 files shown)


debugging/book-library/index.html
```diff
@@ -29,20 +27,24 @@ <h1>Library</h1>
...
           required
         />
         <label for="author">Author: </label>
         <input
-          type="author"
+          type="text"
           class="form-control"
           id="author"
           name="author"
+          minlength="2"
```
illicitonion: How did you decide on these minimum and maximum lengths (for all fields)?

illicitonion (commented):
LGTM, well done! Just one question to consider

debugging/book-library/index.html
RaihanSharif: Tbh, I didn't give it a lot of thought wrt the lengths.

I put 2 as min with the rationale that accidental typos and author name conflicts would be less likely, but that's not a strong argument. 

I had considered a min of 1, as there are single letter pseudonyms and book titles.

The max I set to something very long as size isn't a bottleneck at this scale, but the user still has some sort of limit.

RaihanSharif: After some consideration, I have replaced the min with 1, and the max with 255 as that is usually the limit for database VARCHAR.

illicitonion: Seems reasonable :) I'm less convinced on the max page limit (there are some very _long_ books in this worked), but you have reasonable thoughts on them :)

RaihanSharif: Had to double check my code to see if I accidentally put 255 limit on pages! I'll leave it at 10,000. 

Thank you for your feedback. I have learned to be more thoughful and deliberate with my code.

PR #502, opened 2026-07-30
London | 26-ITP-May | Eyob Zeray | Sprint 2 | Book Library

eyob-tech:
## Changelist

Fixed the following bugs in the Book Library debugging exercise:

1. Website loads but doesn't show any books — missing closing parenthesis in the render() for-loop was causing a syntax error that broke the whole script
2. Error in console when adding a book — submit() was pushing to an undefined library variable instead of myLibrary
3. Title used as author name — Book() was being called with title.value passed twice instead of author.value for the author field
4. Delete button broken — variable was declared as delButton but referenced as delBut, and the event listener used "clicks" instead of "click"
5. Wrong read status saved — the Yes/No display logic was inverted
6. Missing author validation — submit() only checked title and pages for empty values; added a check for author so a book can't be added without one, matching the spec

Tested manually in the browser: books display on load, adding a book with all fields works correctly, missing fields trigger the alert, read status toggles correctly, and delete removes the correct book.
## Learners, PR Template

Self checklist

## Changelist

Fixed the bugs listed in the README:
- Website not showing books: fixed a syntax error in render() (missing closing parenthesis)
- Console error on adding a book: fixed submit() referencing an undefined variable instead of myLibrary
- Title used as author: fixed Book() call to use the author field instead of duplicating the title
- Delete button broken: fixed a variable name mismatch and an incorrect event listener name
- Wrong read status saved: fixed inverted Yes/No logic
- Also added validation so a book can't be submitted with a missing author, per the spec

### DIFF
--- a/debugging/book-library/index.html
+++ b/debugging/book-library/index.html
@@ -3,7 +3,7 @@
   <head>
     <title>Library</title>
     <meta charset="utf-8" />
-<meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.4.1/jquery.min.js"></script>
     <script src="https://cdnjs.cloudflare.com/ajax/libs/popper.js/1.16.0/umd/popper.min.js"></script>
     <script src="https://maxcdn.bootstrapcdn.com/bootstrap/4.4.1/js/bootstrap.min.js"></script>
@@ -49,6 +49,7 @@
           id="pages"
           name="pages"
           min="1"
+          step="1"
           required
         />
         <label class="form-check-label">
--- a/debugging/book-library/index.html
+++ b/debugging/book-library/index.html
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <title> </title>
+    <title>Library</title>
     <meta
       charset="utf-8"
       name="viewport"
@@ -31,7 +31,7 @@
       <div class="form-group">
         <label for="title">Title:</label>
         <input
-          type="title"
+          type="text"
           class="form-control"
           id="title"
           name="title"
@@ -39,7 +39,7 @@
         />
         <label for="author">Author: </label>
         <input
-          type="author"
+          type="text"
           class="form-control"
           id="author"
           name="author"
@@ -51,6 +51,7 @@
           class="form-control"
           id="pages"
           name="pages"
+          min="1"
           required
         />
         <label class="form-check-label">
--- a/debugging/book-library/index.html
+++ b/debugging/book-library/index.html
@@ -1,12 +1,9 @@
 <!DOCTYPE html>
-<html>
+<html lang="en">
   <head>
     <title>Library</title>
-    <meta
-      charset="utf-8"
-      name="viewport"
-      content="width=device-width, initial-scale=1.0"
-    />
+    <meta charset="utf-8" />
+<meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.4.1/jquery.min.js"></script>
     <script src="https://cdnjs.cloudflare.com/ajax/libs/popper.js/1.16.0/umd/popper.min.js"></script>
     <script src="https://maxcdn.bootstrapcdn.com/bootstrap/4.4.1/js/bootstrap.min.js"></script>
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -7,8 +7,8 @@ window.addEventListener("load", function (e) {
 
 function populateStorage() {
   if (myLibrary.length == 0) {
-    let book1 = new Book("Robison Crusoe", "Daniel Defoe", "252", true);
-    let book2 = new Book(
+    const book1 = new Book("Robison Crusoe", "Daniel Defoe", "252", true);
+    const book2 = new Book(
       "The Old Man and the Sea",
       "Ernest Hemingway",
       "127",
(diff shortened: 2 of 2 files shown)


cjyuan (commented):

eyob-tech:
Thanks for the pointer! Went through the general feedback doc and applied everything relevant to book-library:
script.js:
	•	Removed a duplicate render() call that was firing twice on page load
	•	Fixed colliding button IDs between read/delete buttons on different rows
	•	Fixed delete confirmation timing — deletion now completes before the alert shows, not after
	•	Simplified the read-status logic using a ternary operator
	•	Changed non-reassigned lets to const
	•	Renamed variables for clarity (e.g. title → titleInput, changeBut → changeButton)
	•	Switched .innerHTML to .textContent for inserting plain book data
index.html:
	•	Fixed the empty <title> tag
	•	Corrected invalid type="title"/type="author" attributes to type="text"
	•	Added min="1" on the pages input
	•	Fixed an HTML validation error — the charset and viewport meta tags were incorrectly combined into one; split them into two separate <meta> tags
	•	Added lang="en" to the <html> tag
	•	Ran the file through validator.w3.org — no remaining errors (only harmless “Info” notices about optional trailing slashes on void elements)
Let me know if there’s anything else you’d like adjusted!

debugging/book-library/index.html
```diff
@@ -51,6 +48,7 @@ <h1>Library</h1>
           class="form-control"
           id="pages"
           name="pages"
+          min="1"
           required
         />
```
cjyuan, line 54: Page count should be a whole number. How would you update this to accept only a whole number?

debugging/book-library/script.js

debugging/book-library/script.js
```diff
@@ -7,38 +7,44 @@ window.addEventListener("load", function (e) {
...
   if (myLibrary.length == 0) {
-    let book1 = new Book("Robison Crusoe", "Daniel Defoe", "252", true);
-    let book2 = new Book(
+    const book1 = new Book("Robison Crusoe", "Daniel Defoe", "252", true);
+    const book2 = new Book(
       "The Old Man and the Sea",
       "Ernest Hemingway",
       "127",
       true
     );
```
cjyuan, line 16: What type of data should page count be? What allowing the same kind of data to be in different data types is not a good practice?

debugging/book-library/script.js
```diff
@@ -7,38 +7,44 @@ window.addEventListener("load", function (e) {
...
 
 //check the right input from forms and if its ok -> add the new book (object in array)
 //via Book function and start render function
 function submit() {
   if (
-    title.value == null ||
-    title.value == "" ||
-    pages.value == null ||
-    pages.value == ""
+    titleInput.value == null ||
```
cjyuan: When would the `.value` property of a DOM input element be `null`?

debugging/book-library/script.js
```diff
@@ -7,38 +7,44 @@ window.addEventListener("load", function (e) {
...
     return false;
   } else {
-    let book = new Book(title.value, title.value, pages.value, check.checked);
-    library.push(book);
+    let book = new Book(
+      titleInput.value,
+      authorInput.value,
+      pagesInput.value,
+      checkInput.checked
+    );
```
cjyuan, line 49: Using user input directly without sanitization, normalization, or conversion to the proper data type can be dangerous.

How could you prevent "bad input" from ruining the app?

debugging/book-library/script.js
```diff
@@ -51,53 +57,46 @@ function Book(title, author, pages, check) {
...
 function render() {
-  let table = document.getElementById("display");
-  let rowsNumber = table.rows.length;
+  const table = document.getElementById("display");
+  const rowsNumber = table.rows.length;
   //delete old table
-  for (let n = rowsNumber - 1; n > 0; n-- {
+  for (let n = rowsNumber - 1; n > 0; n--) {
     table.deleteRow(n);
   }
```
cjyuan: Clearing a table row by row is not very efficient. Could you look up an efficient approach to clear clear a table?

debugging/book-library/script.js
```diff
@@ -51,53 +57,46 @@ function Book(title, author, pages, check) {
...
-      alert(`You've deleted title: ${myLibrary[i].title}`);
+    const delButton = document.createElement("button");
+    deleteCell.appendChild(delButton);
+    delButton.className = "btn btn-warning";
+    delButton.innerHTML = "Delete";
+    delButton.addEventListener("click", function () {
+      const deletedTitle = myLibrary[i].title;
       myLibrary.splice(i, 1);
       render();
+      alert(`You've deleted title: ${deletedTitle}`);
```
cjyuan: `alert()` is a blocking function call (it would pause execution until the dialog box is closed).

Could you, with the help of AI, explore a way to overcome this shortcoming?

debugging/book-library/index.html
```diff
@@ -1,12 +1,9 @@
...
   <head>
-    <title> </title>
-    <meta
-      charset="utf-8"
-      name="viewport"
-      content="width=device-width, initial-scale=1.0"
-    />
+    <title>Library</title>
+    <meta charset="utf-8" />
+<meta name="viewport" content="width=device-width, initial-scale=1.0" />
```
cjyuan: Indentation is off.

cjyuan (commented):
A good start.

I have some suggestions to further improve the code.

debugging/book-library/index.html
eyob-tech: Added step="1" alongside the existing min="1" on the pages input. Since the form doesn’t use native browser submission (it’s wired up through an onclick handler rather than a real <form>), I also added an explicit check in submit() to reject non-whole numbers and anything below 1, so it’s enforced either way.

debugging/book-library/script.js
eyob-tech: Changed it to const myLibrary = []. It still works fine with .push() and .splice() since those mutate the array in place rather than reassign it — const just blocks something like myLibrary = [] from happening elsewhere by mistake.

debugging/book-library/script.js
eyob-tech: Page count should be a Number, not a string — the original book1/book2 were stored with pages as strings ("252"), and allowing both types made comparisons and formatting inconsistent (e.g. "127" + 1 would concatenate instead of add). Fixed by converting to Number(pages) in the Book() constructor, so every book’s page count is consistently numeric regardless of source.

debugging/book-library/script.js
eyob-tech: Turns out it wouldn’t — .value on a real DOM input always returns a string, even if empty (""), never null. Removed the dead == null checks and kept just the == "" checks, which are the actual condition that can occur.

debugging/book-library/script.js
eyob-tech: Added .trim() on title and author to strip stray whitespace, and Number(pagesInput.value) to ensure pages is always converted to the right type before it’s stored — plus the explicit whole-number/positive check in submit() catches bad numeric input before it reaches the array at all.

debugging/book-library/script.js
eyob-tech: Switched to clearing the whole tbody in one operation instead of looping deleteRow(): tbody.innerHTML = "". Rows are now also inserted directly into the tbody via tbody.insertRow() rather than table.insertRow(1).

debugging/book-library/script.js
eyob-tech: Good call — replaced alert() on delete with a small custom message function instead:
function showMessage(text) {
  const msg = document.createElement("div");
  msg.textContent = text;
  msg.className = "alert alert-info";
  document.body.prepend(msg);
  setTimeout(() => msg.remove(), 2000);
}
It creates a temporary banner at the top of the page instead of a blocking dialog, and removes itself after 2 seconds — so the page stays responsive and doesn’t pause execution.

debugging/book-library/index.html
eyob-tech: Fixed — both meta tags are now aligned with the rest of the <head> block.

cjyuan:
All changes are great and all responses are spot on. Excellent job!

PR #516, opened 2026-08-05
Birmingham | 26-ITP-May | Tobias Amaechina | Sprint 2 | Book-library

Tobias-Amaechina:
Self checklist

## Changelist

Corrected validation logic to ensure title, author, and pages fields are required before adding a book.

Fixed bug where the author field was ignored and replaced with the title value.

Corrected array reference (library → myLibrary) to prevent runtime errors when adding books.

Repaired table‑rendering loop by fixing a missing parenthesis in the row‑deletion logic.

Corrected read/unread status logic so check === true displays “Yes”.

Fixed delete‑button creation:

Corrected variable name (delButton instead of delBut)

Removed accidental reassignment of the button element

Corrected event listener from "clicks" to "click"

Ensured delete action removes the correct book and re-renders the table


cjyuan (commented):

Tobias-Amaechina:
Thank you  CY for the link , it definitely  helped and the HTML has been validated , Thanks

debugging/book-library/index.html
cjyuan, line 58: Why not make this input element to accept only positive whole number?

debugging/book-library/index.html
```diff
@@ -62,13 +63,13 @@ <h1>Library</h1>
           />Read
         </label>
         <input
-          type="submit"
+          type="button"
           value="Submit"
           class="btn btn-primary"
-          onclick="submit();"
+          onclick="handleSubmit();"
```
cjyuan: Suggestion: Look up the trade-off among different approaches for attaching event listener to HTML elements. In particulars,
- Via HTML attribute in HTML
- In JS, using `.addEventListener()`

debugging/book-library/script.js
```diff
@@ -1,103 +1,97 @@
...
-    pages.value == null ||
-    pages.value == ""
-  ) {
-    alert("Please fill all fields!");
-    return false;
-  } else {
-    let book = new Book(title.value, title.value, pages.value, check.checked);
-    library.push(book);
-    render();
+  if (!title || !author || !pages || pages <= 0) {
```
cjyuan, line 43: What type of number should `pages` be? Could it be `NaN`?

cjyuan (commented):
Changes look good.

I only have a few more possible improvements for you to consider.

debugging/book-library/index.html
Tobias-Amaechina: Thank you CJ for the feedback the corrections has been made

debugging/book-library/index.html
```diff
@@ -62,13 +65,13 @@ <h1>Library</h1>
           />Read
         </label>
```
cjyuan, line 66: Did you notice the checkbox is not showing in the form?
The issue is related to Bootstrap 4.4.1 and AI can point out a solution if you show the code to it (and mention the code uses Bootstrap 4.4.1)

Fixing this is optional.

debugging/book-library/script.js
```diff
@@ -1,103 +1,102 @@
...
-  this.check = check;
+  const book = new Book(title, author, pages, readCheck.checked);
+  myLibrary.push(book);
+
+  render();
+  showMessage(`Added "${title}" to your library.`);
 }
 
+// attach listener 
+document.getElementById("submit-btn").addEventListener("click", handleSubmit);
```
cjyuan, line 56: Could consider placing all code that runs once on page load in a single function. For example, you could put it inside the page load callback or create a function named `init()` or `setup()` and call it once when the page loads.

debugging/book-library/script.js
```diff
@@ -1,103 +1,102 @@
...
   }
 }
 
-const title = document.getElementById("title");
-const author = document.getElementById("author");
-const pages = document.getElementById("pages");
-const check = document.getElementById("check");
+// Handle form submission
+ function handleSubmit(e) {
+  e.preventDefault();//stop the form from reloading the page
```
cjyuan, line 37: Note: This method does not stop the page from reloading directly. It stops something else that makes the browser from loading a new page. It's an important method to understand.

debugging/book-library/script.js
cjyuan: Although the HTML is prepared to reject non positive whole number, it is a safer practice to also fully check the input in JS code to ensure it is a positive whole number.

cjyuan (commented):
Changes look good.

PR #519, opened 2026-08-06
Manchester | 26-ITP-May | Abdu Hassen | Sprint 1 | Destructuring

Abduhasen:
## Learners, PR Template

Self checklist

## Changelist

destructuring of codes


Sprint-1/destructuring/exercise-1/exercise.js
```diff
@@ -3,10 +3,10 @@ const personOne = {
   age: 34,
   favouriteFood: "Spinach",
 };
-
+let { name, age, favouriteFood } = personOne;
```
cjyuan, line 6: This would introduce the variables in the "global" scope (scope outside the function).

We could de-structure the parameters too.

Sprint-1/destructuring/exercise-2/exercise.js
```diff
@@ -70,3 +70,26 @@ let hogwarts = [
     occupation: "Teacher",
   },
 ];
+function houseGryffindor(hogwarts) {
+  for (let individualHogwarts of hogwarts) {
```
cjyuan, line 74: - Normally, we can use `const` to declare the loop variable in a for-of or for-in loop because we usually don't need to reassign value to the variable. 

- We could also use de-structure syntax in a for-of loop as `for (const { key1, key2, ... } of objects) { ... }` where `key1`, `key2`, ... are the property names.

Sprint-1/destructuring/exercise-3/exercise.js
```diff
@@ -6,3 +6,26 @@ let order = [
...
+        itemName.padEnd(22) +
+        (itemTotal / 100).toFixed(2)
+    );
+  }
+
+  console.log("---------------------------");
+  console.log("Total: " + (grandTotal / 100).toFixed(2));
+}
+
+valueProperty(order);
```
cjyuan, line 31: If you call this function multiple times  in a row (even with the same argument), you may discover a bug in your code.

PR #521, opened 2026-08-06
London | 26-ITP-May | Dipa Sarker | Sprint 2 | Book Library Project

Dipa-Sarker:
## Learners, PR Template

Self checklist

## Changelist

**Changes** made

- Fixed bugs preventing books from being added correctly to the library.
- Corrected book information handling, including title, author, and read status.
- Improved the rendering of book cards on the page.
- Added/fixed event handling for user interactions.
- Debugged JavaScript issues using browser developer tools and verified the application behaviour.
**Testing**
- Tested adding new books through the form.
- Confirmed that book details display correctly.
- Verified that the read status updates properly.
- Checked that the application works correctly in the browser using Live Server.

**The** changes were tested locally and the application is working as expected.


cjyuan (commented):

Dipa-Sarker:
Thank you for giving me the general feedback file. I made correction according to this.

debugging/book-library/script.js
```diff
@@ -1,17 +1,18 @@
 let myLibrary = [];
```

debugging/book-library/script.js
```diff
@@ -20,28 +21,44 @@ function populateStorage() {
...
 
-const title = document.getElementById("title");
-const author = document.getElementById("author");
-const pages = document.getElementById("pages");
-const check = document.getElementById("check");
+const titleInput = document.getElementById("title"); //rename all variable names
+const authorInput = document.getElementById("author");
+const pagesInput = document.getElementById("pages");
+const checkInput = document.getElementById("check");
+const bookForm = document.getElementById("bookForm");
```
cjyuan: Note: Normal practice is to keep variable/constant declaration at beginning of the file (before function definition).

debugging/book-library/script.js
```diff
@@ -20,28 +21,44 @@ function populateStorage() {
...
+    const book = new Book(title, author, pages, checkInput.checked); //added author.value by
+    // replacing 2nd title.value to author.value
+    myLibrary.push(book); //corrected name for original array library > myLibrary
+    render();
 }
+bookForm.addEventListener("submit", function (event) { //after user submits the book form 
+// this will run submit () function to add book
+  event.preventDefault();
+  submit();
+});
```
cjyuan: Common practice is to keep all the "code that runs once on page load" in the same place (e.g., in a function named `init()` or `setup()`) and then call the function on page load.

Code that setups event listener usually fall into this category.

debugging/book-library/script.js
```diff
@@ -51,53 +68,47 @@ function Book(title, author, pages, check) {
...
-    if (myLibrary[i].check == false) {
-      readStatus = "Yes";
-    } else {
-      readStatus = "No";
-    }
-    changeBut.innerText = readStatus;
-
-    changeBut.addEventListener("click", function () {
+    const changeButton = document.createElement("button"); //correct changeBut > changeButton
+    changeButton.id = i;
```
cjyuan: Is there any need to assign an `id` attribute to this element?
In addition, "0", "1", ..., do not make good ID value.

debugging/book-library/script.js
```diff
@@ -51,53 +68,47 @@ function Book(title, author, pages, check) {
...
+    let deleteButton = document.createElement("button");
+    deleteCell.appendChild(deleteButton);
+    deleteButton.className = "btn btn-warning";
+    deleteButton.textContent = "Delete";
+    deleteButton.addEventListener("click", function () {
+      // correct clicks > click
+      const deletedTitle = myLibrary[i].title; //save the book title which should have to delete
       myLibrary.splice(i, 1);
       render();
+      alert(`You've deleted title: ${deletedTitle}`);
```

cjyuan (commented):
Changes look solid. I just have a few suggestions.

debugging/book-library/script.js
Dipa-Sarker: we can use const instead of let.

debugging/book-library/script.js
Dipa-Sarker: okay, moved all the variables at the very beginning.

debugging/book-library/script.js
Dipa-Sarker: ok, we don't need button id, because we have event listener to response with change button. Besides, only id ="0", "1",...etc don't mean anything. I deleted this line.

debugging/book-library/script.js
Dipa-Sarker: Ok, I got your point. I did two things. 

1. Use confirm() before deleting.
2. Put a normal message element in the HTML instead of using alert() after deletion.

debugging/book-library/script.js
```diff
@@ -1,10 +1,31 @@
...
+
+const myLibrary = []; //use const to avoid reassignment
 
-window.addEventListener("load", function () {
+/*window.addEventListener("load", function () {
   //delete e as it is never used
   populateStorage();
   render();
-});
+}); */
```
cjyuan, line 15: In practice, it is better to remove any unused code instead of wrap them in a comment in order to keep the code clean.

debugging/book-library/script.js
```diff
@@ -98,17 +106,29 @@ function render() {
...
-      const deletedTitle = myLibrary[i].title; //save the book title which should have to delete
-      myLibrary.splice(i, 1);
-      render();
-      alert(`You've deleted title: ${deletedTitle}`);
-    });
+      const deletedTitle = myLibrary[i].title; // store the title which book will delete
+
+       const confirmed = confirm( //use confirm () to handle popup message for ok/cancel
+      `Are you sure you want to delete "${deletedTitle}"?`
+       );
```
cjyuan, line 122: Indentation is off.

cjyuan (commented):
Changes look good. Well done.

Dipa-Sarker:
Thank you for your feedback!

PR #523, opened 2026-08-07
London | 26-ITP-May | Martin Mwaka | Sprint 2 | Book library

Temceo:
Self checklist

## Changelist
Debug and improve functionality of the library app

### DIFF
--- a/debugging/book-library/index.html
+++ b/debugging/book-library/index.html
@@ -38,8 +38,7 @@
           </div>
           <div>
             <label for="pages">Pages:</label>
-            <input type="text" class="form-control" id="pages" name="pages" required inputmode="numeric"
-              pattern="[0-9]*" autocomplete="off">
+            <input type="number" class="form-control" id="pages" name="pages" min="1" required>
             <span class="error error-pages"></span>
           </div>
 
--- a/debugging/book-library/index.html
+++ b/debugging/book-library/index.html
@@ -13,17 +13,6 @@
 </head>
 
 <body>
-  <template id="book-template">
-    <thead>
-      <tr>
-        <th>Title</th>
-        <th>Author</th>
-        <th>Pages</th>
-        <th>Read</th>
-        <th>Delete</th>
-      </tr>
-    </thead>
-  </template>
   <main>
     <header>
       <div class="jumbotron text-center">
@@ -39,29 +28,44 @@
         <form class="form-group" id="add-book">
           <div>
             <label for="title">Title:</label>
-            <input type="text" class="form-control" id="title" name="title">
+            <input type="text" class="form-control" id="title" name="title" required>
             <span class="error error-title"></span>
           </div>
           <div>
             <label for="author">Author: </label>
-            <input type="text" class="form-control" id="author" name="author">
+            <input type="text" class="form-control" id="author" name="author" required>
             <span class="error error-author"></span>
           </div>
           <div>
             <label for="pages">Pages:</label>
-            <input type="number" class="form-control" id="pages" name="pages">
+            <input type="text" class="form-control" id="pages" name="pages" required inputmode="numeric"
+              pattern="[0-9]*" autocomplete="off">
             <span class="error error-pages"></span>
           </div>
 
-          <label class="form-check-label">
-            <input type="checkbox" class="form-check-input" id="check" value="">Read
+          <label class="form-check-label" for="check">
+            <input type="checkbox" class="form-check-input" id="check" name="check" value="read">Read
           </label>
 
           <input type="submit" value="Submit" class="btn btn-primary" id="submit-button">
         </form>
       </div>
     </header>
-    <div class="display-books"></div>
+    <table id="book-table">
+      <thead>
+        <tr>
+          <th>Title</th>
+          <th>Author</th>
+          <th>Pages</th>
+          <th>Read</th>
+          <th>Delete</th>
+        </tr>
+      </thead>
+      <tbody id="display-books">
+
+      </tbody>
+    </table>
+
 
   </main>
 
--- a/debugging/book-library/index.html
+++ b/debugging/book-library/index.html
@@ -1,96 +1,63 @@
 <!DOCTYPE html>
-<html>
-  <head>
-    <title> </title>
-    <meta
-      charset="utf-8"
-      name="viewport"
-      content="width=device-width, initial-scale=1.0"
-    />
-    <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.4.1/jquery.min.js"></script>
-    <script src="https://cdnjs.cloudflare.com/ajax/libs/popper.js/1.16.0/umd/popper.min.js"></script>
-    <script src="https://maxcdn.bootstrapcdn.com/bootstrap/4.4.1/js/bootstrap.min.js"></script>
-    <link
-      rel="stylesheet"
-      href="https://maxcdn.bootstrapcdn.com/bootstrap/4.4.1/css/bootstrap.min.css"
-    />
-    <link rel="stylesheet" type="text/css" href="style.css" />
-  </head>
+<html lang="en">
 
-  <body>
-    <div class="jumbotron text-center">
-      <h1>Library</h1>
-      <p>Add books to your virtual library</p>
+<head>
+  <title> </title>
+  <meta charset="utf-8" name="viewport" content="width=device-width, initial-scale=1.0" />
+  <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.4.1/jquery.min.js"></script>
+  <script src="https://cdnjs.cloudflare.com/ajax/libs/popper.js/1.16.0/umd/popper.min.js"></script>
+  <script src="https://maxcdn.bootstrapcdn.com/bootstrap/4.4.1/js/bootstrap.min.js"></script>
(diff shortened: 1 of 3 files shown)


cjyuan (commented):

Temceo:
I have reviewed the general feedback and improved the code

debugging/book-library/index.html
```diff
@@ -1,96 +1,71 @@
...
+          </div>
+          <div>
+            <label for="author">Author: </label>
+            <input type="text" class="form-control" id="author" name="author">
+            <span class="error error-author"></span>
+          </div>
+          <div>
+            <label for="pages">Pages:</label>
+            <input type="number" class="form-control" id="pages" name="pages">
+            <span class="error error-pages"></span>
```
cjyuan, line 42: Why not make this input element reject unwanted input values?

debugging/book-library/script.js
```diff
@@ -1,103 +1,172 @@
...
+const addBookForm = document.getElementById("add-book");
+const submitBtn = document.getElementById("submit-button");
+const titleInput = document.getElementById("title");
+const authorInput = document.getElementById("author");
+const pagesInput = document.getElementById("pages");
+const checkInput = document.getElementById("check");
+const displayBooks = document.querySelector(".display-books");
+const bookEntryTab = document.getElementById("demo");
+
 let myLibrary = [];
```

debugging/book-library/script.js
```diff
@@ -1,103 +1,172 @@
...
-    let book2 = new Book(
-      "The Old Man and the Sea",
-      "Ernest Hemingway",
-      "127",
-      true
+  if (myLibrary.length === 0) {
+    myLibrary.push(new Book("Robison Crusoe", "Daniel Defoe", "252", true));
+    myLibrary.push(
+      new Book("The Old Man and the Sea", "Ernest Hemingway", "127", true)
     );
```
cjyuan: Why represent page count as a string?

debugging/book-library/script.js
```diff
@@ -1,103 +1,172 @@
...
-    alert("Please fill all fields!");
-    return false;
-  } else {
-    let book = new Book(title.value, title.value, pages.value, check.checked);
-    library.push(book);
+// form submission event listeners - add new books
+addBookForm.addEventListener("submit", (event) => {
+  event.preventDefault();
+  processEntries();
+});
```
cjyuan: To make locating all the "code that is to be executed once on page load" easier, a common practice is to keep the code in the same place. For example, inside the page's onload callback function.

debugging/book-library/script.js
```diff
@@ -1,103 +1,172 @@
...
+    if (value === "") {
+      field.error.textContent = field.message;
+      isValid = false;
+    } else {
+      field.error.textContent = "";
+    }
+  });
+
+  if (isValid) {
+    let book = new Book(title.value, author.value, pages.value, check.checked);
```
cjyuan: - The raw values used here could be different from those checked on line 75.

- `pages.value` could be a value like `"3e2"` or `"12.345"`. Should page count be assigned values like these?

debugging/book-library/script.js
```diff
@@ -1,103 +1,172 @@
...
+
+  const thead = document.createElement("thead");
+  const headerRow = document.createElement("tr");
+  ["Title", "Author", "Pages", "Read", "Delete"].forEach((label) => {
+    const th = document.createElement("th");
+    th.textContent = label;
+    headerRow.appendChild(th);
+  });
+  thead.appendChild(headerRow);
+  table.appendChild(thead);
```
cjyuan: Why dynamically create the static table its header rows every time `render()` is called? Why not keep them in `index.html` as HTML code?

debugging/book-library/script.js
Temceo: Hi, can I please clarify on this point.  The addForm event listener only runs each time a new book is added and the processEntries() function validates the inputs before adding the new book. The code that is run on page load is the populateStorage() function which calls the render() function to show all existing books in the library

debugging/book-library/index.html
Temceo: I have add bespoke validation that prevents submission unless a whole number is added and an error message is shown to the user until the correct input is added

debugging/book-library/script.js
Temceo: I have now declared this using const

debugging/book-library/script.js
Temceo: I have changed these to numbers

debugging/book-library/script.js
Temceo: I have updated the event listeners for window load and add book so hopefully now it is much clearer

debugging/book-library/script.js
Temceo: I have added validation to check a whole number is entered before page count is assigned a value

debugging/book-library/script.js
Temceo: I have reverted to html code for this

debugging/book-library/script.js
```diff
@@ -80,48 +110,24 @@ function processEntries() {
...
+  }
+
   if (isValid) {
-    let book = new Book(title.value, author.value, pages.value, check.checked);
+    let book = new Book(
+      titleInput.value,
+      authorInput.value,
+      pagesInput.value,
+      checkInput.checked
+    );
```
cjyuan, line 127: When `titleInput.value` and `authorInput.value` have leading and trailing space, they can pass the validation but the code here would store the raw input.

Similarly, when `pagesInput.value` is `"0010"`, the value can pass the validation but the code here will store `"0010"` (instead of `10`) as the page count.

A common approach to deal with user input is to first store the preprocessed input in variables,

debugging/book-library/script.js
```diff
@@ -56,17 +46,57 @@ const fields = {
     message: "Enter page numbers (whole numbers only)",
   },
 };
+// validate that pages input is a whole number
+function validatePagesInput() {
+  const rawValue = pagesInput.value.trim();
+  const isValidPages =
+    rawValue !== "" && /^\d+$/.test(rawValue) && Number(rawValue) >= 1;
```
cjyuan: You could also just convert the input to a number first and then just check the number. This way, we don't need code to check the string input.

debugging/book-library/index.html
```diff
@@ -39,29 +28,44 @@ <h1>Library</h1>
...
             <label for="author">Author: </label>
-            <input type="text" class="form-control" id="author" name="author">
+            <input type="text" class="form-control" id="author" name="author" required>
             <span class="error error-author"></span>
           </div>
           <div>
             <label for="pages">Pages:</label>
-            <input type="number" class="form-control" id="pages" name="pages">
+            <input type="text" class="form-control" id="pages" name="pages" required inputmode="numeric"
+              pattern="[0-9]*" autocomplete="off">
```
cjyuan: - Changing this input to `type="text"` affects accessibility; screen reader announces input of type "text" differently from input of type "number". 

- `<input type="number" class="form-control" id="pages" name="pages" required min="1">` also works. (Note: by default, `step="1"` is implied).

debugging/book-library/script.js
cjyuan: 1. Basically all statements written in the top-level scope are code that would be executed once when the HTML page first loaded.
  - Code that attach a callback to a DOM element once usually falls into this category.

2. Typically we order the code in a file the following manner:
```
  variable declarations
  
  function definitions
  
  code to be executed (code to be executed once is placed here)
```

3.  Alternatively, we can place all the code to be executed once on page load inside a function (e.g., `init()`), and then just call the function.

---
What you did works. 

The code that manages form validation could be better organised but that's out of the scope of this exercise.

debugging/book-library/script.js
```diff
@@ -80,48 +110,24 @@ function processEntries() {
...
+      titleInput.value,
+      authorInput.value,
+      pagesInput.value,
+      checkInput.checked
+    );
     myLibrary.push(book);
     bookEntryTab.classList.remove("show");
     render();
-    addBookForm.reset();
+    addBook.reset();
```
cjyuan: Why not use the name `addBookForm`? It makes it clearer that the variable refers to a form element.

debugging/book-library/script.js
Temceo: I have created variables for all 3.  For titleInput and authorInput the input values are being trimmed before insertion into the book class.  For pagesInput I have added validation to prevent entry of leading zeros and negative numbers

debugging/book-library/index.html
Temceo: I have reverted this to type="number"

debugging/book-library/script.js
Temceo: I have reverted this to addBookForm

debugging/book-library/script.js
```diff
@@ -115,16 +115,20 @@ function processEntries(event) {
...
+    const trimmedAuthor = authorInput.value.trim();
+    const trimmedPages = pages.value.trim();
+
     let book = new Book(
-      titleInput.value,
-      authorInput.value,
-      pagesInput.value,
+      trimmedTitle,
+      trimmedAuthor,
+      trimmedPages,
```
cjyuan, line 125: Should the `.pages` property of a book be a number or a string? For this particular simple app it may not matter but it is safer to consistently represent a kind of data using same data type.

cjyuan (commented):
Changes look good.

PR #524, opened 2026-08-07
Glasgow | 26-ITP-May | Sandani Kannangara | Sprint 2 | Fix library debugging

SandzSoft:
Self checklist

## Changelist

- Fixed books not rendering on page load.
- Fixed errors when adding new books.
- Fixed author field using title value.
- Fixed delete button functionality.
- Fixed read status handling.
- Added validation for required fields.
- Added form reset after adding books.
- Refactored DOM handling to use tbody.
- Improved variable naming and readability.

### DIFF
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -37,8 +37,8 @@ function submit() {
     alert("Please fill all fields!");
     return false;
   } else {
-    let book = new Book(title.value, title.value, pages.value, check.checked);
-    library.push(book);
+    let book = new Book(title.value, author.value, pages.value, check.checked);
+    myLibrary.push(book);
     render();
   }
 }
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -30,8 +30,6 @@ function populateStorage() {
   }
 }
 
-
-
 function addBook() {
   const title = titleInput.value.trim();
   const author = authorInput.value.trim();
@@ -80,7 +78,7 @@ function render() {
 
     const changeButton = document.createElement("button");
     changeButton.className = "btn btn-success";
-    changeButton.textContent = myLibrary[i].isRead ? "No" : "Yes";
+    changeButton.textContent = myLibrary[i].isRead ? "Yes" : "No";
 
     wasReadCell.appendChild(changeButton);
 
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -1,8 +1,18 @@
-let myLibrary = [];
+const myLibrary = [];
+
+const titleInput = document.getElementById("title");
+const authorInput = document.getElementById("author");
+const pagesInput = document.getElementById("pages");
+const isReadInput = document.getElementById("isRead");
+const bookForm = document.getElementById("book-form");
 
 window.addEventListener("load", function () {
   populateStorage();
   render();
+  bookForm.addEventListener("submit", function (event) {
+    event.preventDefault();
+    addBook();
+  });
 });
 
 function populateStorage() {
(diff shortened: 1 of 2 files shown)


cjyuan (commented):

cjyuan (commented):
According to https://validator.w3.org/, there are errors in your `index.html`. Could you fix these errors?

debugging/book-library/script.js
```diff
@@ -1,103 +1,109 @@
 let myLibrary = [];
```
cjyuan: Could we declare `myLibrary` in a way that prevents it from being accidentally reassigned?

debugging/book-library/script.js
```diff
@@ -1,103 +1,109 @@
...
-    return false;
-  } else {
-    let book = new Book(title.value, title.value, pages.value, check.checked);
-    library.push(book);
-    render();
+const titleInput = document.getElementById("title");
+const authorInput = document.getElementById("author");
+const pagesInput = document.getElementById("pages");
+const isReadInput = document.getElementById("isRead");
+const bookForm = document.getElementById("book-form");
```
cjyuan: Common practice is to declare all variables before function definitions.

debugging/book-library/script.js
```diff
@@ -1,103 +1,109 @@
...
+
+bookForm.addEventListener("submit", function (event) {
+  event.preventDefault();
+  addBook();
+});
```
cjyuan: Why not place all "code that runs once on page load" in one place. For example, inside the on page load callback function.

debugging/book-library/script.js
```diff
@@ -1,103 +1,109 @@
...
+
+  clearForm();
+  render();
+}
+
+function clearForm() {
+  titleInput.value = "";
+  authorInput.value = "";
+  pagesInput.value = "";
+  isReadInput.checked = false;
```
cjyuan: You could also use the form's built-in `.reset()` method to reset the form.

debugging/book-library/script.js
```diff
@@ -1,103 +1,109 @@
...
+
+    deleteCell.appendChild(deleteButton);
+
+    deleteButton.addEventListener("click", function () {
+      const deletedTitle = myLibrary[i].title;
+
       myLibrary.splice(i, 1);
       render();
+
+      alert(`You've deleted title: ${deletedTitle}`);
```

debugging/book-library/script.js
SandzSoft: Changed myLibrary from let to const

debugging/book-library/script.js
SandzSoft: Moved the DOM variable declarations before the function definitions.

debugging/book-library/script.js
SandzSoft: Moved the one-time form event-listener setup into the page-load callback.

debugging/book-library/script.js
SandzSoft: Used bookForm.reset() to reset the form.

debugging/book-library/script.js
SandzSoft: For the optional alert() suggestion, I used setTimeout(..., 0) so the UI can update before the alert is displayed.

SandzSoft:
> According to https://validator.w3.org/, there are errors in your `index.html`. Could you fix these errors?

Fixed the HTML validation errors and verified index.html with the W3C validator. It now passes validation.

SandzSoft:
Thanks for the helpful feedback! I've fixed the issues and learned a lot from the suggestions.

cjyuan (commented):
Changes look good, but (because you deleted something needed by Bootstrap) the form no longer works properly. Could you fix this issue?

SandzSoft:
> Changes look good, but (because you deleted something needed by Bootstrap) the form no longer works properly. Could you fix this issue?

Thanks for pointing this out! I fixed the HTML validation issues and restored the required Bootstrap JavaScript dependencies.

cjyuan:
All good now.

PR #525, opened 2026-08-07
London | 26-ITP-May | Chun Yan Wong | Sprint 2 | Book library

cywong-dev:
## Learners, PR Template

Self checklist

## Changelist

This PR provides the answer for the coursework book library

### DIFF
--- a/debugging/book-library/index.html
+++ b/debugging/book-library/index.html
@@ -1,11 +1,9 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <title> </title>
-    <meta
-      charset="utf-8"
-      name="viewport"
-      content="width=device-width, initial-scale=1.0"
+    <title>Virtual Library</title>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0"
     />
     <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.4.1/jquery.min.js"></script>
     <script src="https://cdnjs.cloudflare.com/ajax/libs/popper.js/1.16.0/umd/popper.min.js"></script>
@@ -31,7 +29,7 @@
       <div class="form-group">
         <label for="title">Title:</label>
         <input
-          type="title"
+          type="text"
           class="form-control"
           id="title"
           name="title"
@@ -39,7 +37,7 @@
         />
         <label for="author">Author: </label>
         <input
-          type="author"
+          type="text"
           class="form-control"
           id="author"
           name="author"
--- a/debugging/book-library/index.html
+++ b/debugging/book-library/index.html
@@ -64,9 +64,9 @@
         <input
           type="submit"
           value="Submit"
+          id="submitBtn"
           class="btn btn-primary"
-          onclick="submit();"
-        />
+          />
       </div>
     </div>
 
--- a/debugging/book-library/index.html
+++ b/debugging/book-library/index.html
@@ -51,7 +51,6 @@
           name="pages"
           min="1"
           step="1"
-          oninput="this.value = this.value.replace(/[^0-9]/g, '').replace(/^0+/, '');"
           required
         />
         <label class="form-check-label">
--- a/debugging/book-library/index.html
+++ b/debugging/book-library/index.html
@@ -80,7 +80,7 @@
           <th></th>
         </tr>
       </thead>
-      <tbody>
+      <tbody id="tableBody">
         <tr>
           <td></td>
           <td></td>
--- a/debugging/book-library/index.html
+++ b/debugging/book-library/index.html
@@ -49,6 +49,9 @@
           class="form-control"
           id="pages"
           name="pages"
+          min="1"
+          step="1"
+          oninput="this.value = this.value.replace(/[^0-9]/g, '').replace(/^0+/, '');"
           required
         />
         <label class="form-check-label">
--- a/debugging/book-library/index.html
+++ b/debugging/book-library/index.html
@@ -26,6 +26,7 @@
     </button>
 
     <div id="demo" class="collapse">
+     <form id="book-form"> 
       <div class="form-group">
         <label for="title">Title:</label>
         <input
@@ -68,6 +69,7 @@
           class="btn btn-primary"
           />
       </div>
+     </form>  
     </div>
 
     <table class="table" id="display">
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -54,7 +54,7 @@ function render() {
   let table = document.getElementById("display");
   let rowsNumber = table.rows.length;
   //delete old table
-  for (let n = rowsNumber - 1; n > 0; n-- {
+  for (let n = rowsNumber - 1; n > 0; n--) {
     table.deleteRow(n);
   }
   //insert updated row and cells
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -76,7 +76,7 @@ function render() {
     changeBut.className = "btn btn-success";
     wasReadCell.appendChild(changeBut);
     let readStatus = "";
-    if (myLibrary[i].check == false) {
+    if (myLibrary[i].check == true) {
       readStatus = "Yes";
     } else {
       readStatus = "No";
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -94,7 +94,7 @@ function render() {
     deleteCell.appendChild(delBut);
     delBut.className = "btn btn-warning";
     delBut.innerHTML = "Delete";
-    delBut.addEventListener("clicks", function () {
+    delBut.addEventListener("click", function () {
       alert(`You've deleted title: ${myLibrary[i].title}`);
       myLibrary.splice(i, 1);
       render();
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -48,8 +48,8 @@ function render() {
   let table = document.getElementById("display");
   let rowsNumber = table.rows.length;
   //delete old table
-  for (let n = rowsNumber - 1; n > 0; n--) {
-    table.deleteRow(n);
+  while (table.rows.length > 1) {
+    table.deleteRow(1);
   }
   //insert updated row and cells
   let length = myLibrary.length;
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -3,6 +3,12 @@ const myLibrary = [];
 window.addEventListener("load", function (e) {
   populateStorage();
   render();
+
+  // Connect the form submit event listener
+  const formElement = document.getElementById("book-form");
+  if (formElement) {
+    formElement.addEventListener("submit", submit);
+  }  
 });
 
 function populateStorage() {
@@ -49,8 +55,12 @@ function createStatusContainer() {
 
 //check the right input from forms and if its ok -> add the new book (object in array)
 //via Book function and start render function
-function submit() {
-
+function submit(event) {
+  // Prevent form submission from reloading the page
+  if (event) {
+    event.preventDefault();
+  }
+  
   // Date : 9/8/2026
   // Purpose : Fix the problem "3. It uses the title name as the author name"
   // Change : replace the second title with author.value to fix the bug 
@@ -78,6 +88,17 @@ function submit() {
   let book = new Book(titleValue, authorValue, pagesValue, checkElement.checked);
 
   myLibrary.push(book);
+    // Clear inputs on successful submit
+  if (event && event.target) {
+    event.target.reset();
+  } else {
+    titleElement.value = "";
+    authorElement.value = "";
+    pagesElement.value = "";
+    checkElement.checked = false;
+  }
+
+  displayStatusMessage(`Added "${titleValue}" to your library!`, false);
   render();
   
 }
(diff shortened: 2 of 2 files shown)


cjyuan (commented):
- The checkboxes in the PR description are not yet properly marked up in Markdown syntax.

- Can you check if any of this general feedback can help you further improve your code?

cywong-dev:
Yes, changed. Please kindly check?

debugging/book-library/script.js
```diff
@@ -8,15 +8,13 @@ window.addEventListener("load", function (e) {
...
-    let book2 = new Book(
-      "The Old Man and the Sea",
-      "Ernest Hemingway",
-      "127",
-      true
-    );
-    myLibrary.push(book1);
-    myLibrary.push(book2);
-    render();
+    let book2 = new Book("The Old Man and the Sea","Ernest Hemingway",127,true);
```
cjyuan, line 17: `book1.pages` is a string and `book2.pages` is a number. Better and safer practice is to present the same kind of data using the same data type.

debugging/book-library/script.js
```diff
@@ -29,17 +27,22 @@ const check = document.getElementById("check");
...
-    render();
+    return;
+  } 
+
+
+  // Changed by Chun Yan Wong
+  // Date : 9/8/2026
+  // Purpose : Fix the problem "3. It uses the title name as the author name"
+  // Change : replace the second title with author.value to fix the bug 
+  let book = new Book(title.value, author.value, pages.value, check.checked);
```
cjyuan: - What type of data should `book.pages` be? What type of data is `pages.value`? What invalid "page count" value should be rejected?

- The code checks the trimmed value but it uses the raw (untrimmed) values.

debugging/book-library/script.js
```diff
@@ -29,17 +27,22 @@ const check = document.getElementById("check");
...
-    return false;
-  } else {
-    let book = new Book(title.value, title.value, pages.value, check.checked);
-    library.push(book);
-    render();
+    return;
+  } 
+
+
+  // Changed by Chun Yan Wong
```
cjyuan: Why keep this remark in the comment? Who else is changing this file?

debugging/book-library/script.js
```diff
@@ -54,50 +57,55 @@ function render() {
   let table = document.getElementById("display");
   let rowsNumber = table.rows.length;
   //delete old table
-  for (let n = rowsNumber - 1; n > 0; n-- {
-    table.deleteRow(n);
+  while (table.rows.length > 1) {
+    table.deleteRow(1);
   }
```
cjyuan: Clearing a table by deleting rows one at a time a time is not every efficient.

Could you research a more efficient way to clear a table element?

debugging/book-library/script.js
```diff
@@ -54,50 +57,55 @@ function render() {
...
+    // Changed by Chun Yan Wong
+    // Date : 9/8/2026
+    // Purpose : Fix the problem "4. Delete button is broken"
+    // Change : Fix the incorrect nanmes "delBut" button and "clicks" event to the correct one "delButton" and "click" respectively     
+    delButton.className = "btn btn-warning";
+    delButton.innerText = "Delete";
+    delButton.addEventListener("click", () => {
       alert(`You've deleted title: ${myLibrary[i].title}`);
       myLibrary.splice(i, 1);
       render();
```

In addition, `alert()` is a blocking function call. As a result, invoking it prevents the browser from updating the UI until the dialog is dismissed. 

Suggestion: Research for approaches that allows the UI to update before displaying the `alert` dialog.

cjyuan (commented):

cywong-dev:
> According to https://validator.w3.org/, there are errors in your `index.html`. Can you fix these errors?

Fixed. Please kindly check?

debugging/book-library/script.js
cywong-dev: Yes, it is. Fixed. Please kindly check?

debugging/book-library/script.js
cywong-dev: Removed. Please kindly check?

debugging/book-library/script.js
cywong-dev: Yes, it is fixed. Please kindly check?

debugging/book-library/script.js
cywong-dev: Fixed. Please kindly check?

debugging/book-library/script.js
cywong-dev: Added a helper function to display message in status division instead of using the alert function

debugging/book-library/script.js
```diff
@@ -25,22 +23,51 @@ const author = document.getElementById("author");
...
-    return false;
-  } else {
-    let book = new Book(title.value, title.value, pages.value, check.checked);
-    library.push(book);
-    render();
+    displayStatusMessage("Please fill all fields!", true);
+    return;
   }
+
+  let book = new Book(titleValue, authorValue, pagesValue, check.checked);
```
cjyuan: `pagesValue` is a string, which may hold values like `"000080"`, `"3e2"`, or `"12.345"`.

debugging/book-library/script.js
```diff
@@ -52,52 +79,65 @@ function Book(title, author, pages, check) {
...
+
   //delete old table
-  for (let n = rowsNumber - 1; n > 0; n-- {
-    table.deleteRow(n);
-  }
-  //insert updated row and cells
-  let length = myLibrary.length;
-  for (let i = 0; i < length; i++) {
-    let row = table.insertRow(1);
+  table.innerHTML = "";
```
cjyuan: You could also clear only `<tbody>`. This way, you don't need to recreate the header row.

debugging/book-library/script.js
```diff
@@ -25,22 +23,51 @@ const author = document.getElementById("author");
 const pages = document.getElementById("pages");
 const check = document.getElementById("check");
```
cjyuan: Could rename these variable to indicate that they store DOM elements.

debugging/book-library/script.js

debugging/book-library/index.html
cjyuan, line 56: Why not make this element to accept only positive whole number?

cjyuan: Could you use AI to find out the tradeoff between attaching event listener in HTML and in JavaScript?

cywong-dev: Yes, changed the index.html to add validation. Please kindly check?

debugging/book-library/script.js
cywong-dev: Changed the index.html to validate and accept only positive whole number. Please kindly check?

debugging/book-library/index.html
```diff
@@ -51,6 +49,9 @@ <h1>Library</h1>
           class="form-control"
           id="pages"
           name="pages"
+          min="1"
+          step="1"
+          oninput="this.value = this.value.replace(/[^0-9]/g, '').replace(/^0+/, '');"
           required
```
cjyuan, line 55: - The JS code works to certain extent but it introduces some unusual behavior to the input field. For example, when I pressed '1', '2', '3', 'e', the input field was cleared.

- Adding `min="1"`, `step="1"` and `required` would be enough if the input element is inside a `<form>`. When the input elements are in a `<form>` , the constraints would have been checked by the browser when the user submits the form (clicks the "Submit" button).

debugging/book-library/script.js
cywong-dev: Yes, they are renamed in the script.js.

debugging/book-library/script.js
cywong-dev: Yes, changing "let" to "const" will fix it and it is fixed.

debugging/book-library/script.js
cjyuan: This would lead to
- New books store page count as strings.
- The books created in `populateStorage()` store page count as numbers. 

Could you find a way to represent same kind of value using the same data type?

---
Also, even if the HTML code is already prepared to accept only input containing digits, it is still a safe practice to check the input again in JS.

debugging/book-library/index.html
cywong-dev: Yes, revised.

debugging/book-library/script.js
cywong-dev: Yes, added validation in javascript.

debugging/book-library/index.html
cywong-dev: From AI, in modern web development (vanilla JavaScript or framework-free code), attaching event listeners inside JavaScript via addEventListener is overwhelmingly preferred. It ensures your code is secure, scalable, and easy to maintain. I changed the html and javascript for it

debugging/book-library/script.js
cywong-dev: Yes, changed. Please kindly check?

debugging/book-library/index.html
cjyuan: If the input elements are not placed inside a `<form>` element, the browser won't check the input against the specified constrants when the user clicks the submit button.

cywong-dev: Yes, fixed.Please kindly check?

debugging/book-library/script.js
```diff
@@ -1,46 +1,93 @@
...
-const pages = document.getElementById("pages");
-const check = document.getElementById("check");
+const formElement = document.getElementById("book-form");
+const titleElement = document.getElementById("title");
+const authorElement = document.getElementById("author");
+const pagesElement = document.getElementById("pages");
+const checkElement = document.getElementById("check");
+
+const submitBtn = document.getElementById("submitBtn");
+submitBtn.addEventListener("click", submit);
```
cjyuan: Did you notice `submit` is assigned as callback function twice? Once to this button and once to the form?

debugging/book-library/script.js
```diff
@@ -1,46 +1,93 @@
...
+    return;
+  }
+
+  const positiveIntegerRegex = /^[1-9]\d*$/;
+  if (!positiveIntegerRegex.test(pagesValue)) {
+    displayStatusMessage("Pages must be a positive whole number!", true);
+    return;
   }
+  
+  let book = new Book(titleValue, authorValue, pagesValue, checkElement.checked);
```
cjyuan: - A value such as `"0123"` would be rejected by this check event though technically it is a positive integer.

- Why not convert `pageValues` to a value of type number first? It is easier to check if a number is a positive integer.

debugging/book-library/script.js
```diff
@@ -1,46 +1,93 @@
...
+
+  myLibrary.push(book);
+    // Clear inputs on successful submit
+  formElement.reset();
```
cjyuan, line 90: Good catch fixing the bug and nice touch.

debugging/book-library/script.js
```diff
@@ -52,52 +99,47 @@ function Book(title, author, pages, check) {
...
+  
     let titleCell = row.insertCell(0);
     let authorCell = row.insertCell(1);
     let pagesCell = row.insertCell(2);
     let wasReadCell = row.insertCell(3);
     let deleteCell = row.insertCell(4);
+    
     titleCell.innerHTML = myLibrary[i].title;
     authorCell.innerHTML = myLibrary[i].author;
     pagesCell.innerHTML = myLibrary[i].pages;
```
cjyuan: Have you found out the differences among assigning text to `.innerHTML`, `innerText`, and `textContent` of a DOM element?

debugging/book-library/index.html
cjyuan, line 64: Did you notice the checkbox is not showing in the form?
The issue is related to Bootstrap 4.4.1 and AI can quickly point out a solution if you show the code to it (and mention the code uses Bootstrap 4.4.1)

Fixing this is optional.

cjyuan (commented):
Changes look good. Almost done.

debugging/book-library/index.html
cywong-dev: Thanks! Since it's optional, I'll leave it as-is for now to keep things simple and focus on other coursework.

debugging/book-library/script.js
cywong-dev: Yes, I checked them with AI and know the differences now.

debugging/book-library/script.js
cywong-dev: Thanks

debugging/book-library/script.js
cywong-dev: Fixed. Please kindly check?

debugging/book-library/script.js
cywong-dev: Fixed. Please kindly check?

debugging/book-library/script.js
```diff
@@ -27,7 +27,7 @@ const pagesElement = document.getElementById("pages");
 const checkElement = document.getElementById("check");
 
 const submitBtn = document.getElementById("submitBtn");
```
cjyuan: Is the variable `submitBtn` still needed in the code?

debugging/book-library/script.js
cjyuan: Since you didn't change `innerHTML`, could you describe a user input that could potentially ruin the display (to show that you understand its potential danger)?

cywong-dev: Malicious Script Execution (Cross-Site Scripting / XSS)
While modern browsers generally do not execute simple <script> tags inserted directly via innerHTML, an attacker can easily bypass this using HTML event handlers on elements like  or <svg>.
I have updated the script for it too.

debugging/book-library/script.js
cywong-dev: No, it is not needed. I removed it too.

cjyuan (commented):
All good now.

PR #533, opened 2026-08-09
London | 26-ITP-May | Damilola Odumosu | Sprint 2 | Book Library

d-odumosu:
## Learners, PR Template

Self checklist

### DIFF
--- a/debugging/book-library/index.html
+++ b/debugging/book-library/index.html
@@ -1,12 +1,9 @@
 <!DOCTYPE html>
-<html>
+<html lang="">
   <head>
-    <title> </title>
-    <meta
-      charset="utf-8"
-      name="viewport"
-      content="width=device-width, initial-scale=1.0"
-    />
+    <title>Book Library</title>
+    <meta charset="utf-8"/>
+    <meta name="viewport" content="width=device-width, initial-scale=1.0"/>
     <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.4.1/jquery.min.js"></script>
     <script src="https://cdnjs.cloudflare.com/ajax/libs/popper.js/1.16.0/umd/popper.min.js"></script>
     <script src="https://maxcdn.bootstrapcdn.com/bootstrap/4.4.1/js/bootstrap.min.js"></script>
@@ -31,7 +28,7 @@
       <div class="form-group">
         <label for="title">Title:</label>
         <input
-          type="title"
+          type="text"
           class="form-control"
           id="title"
           name="title"
@@ -39,7 +36,7 @@
         />
         <label for="author">Author: </label>
         <input
-          type="author"
+          type="text"
           class="form-control"
           id="author"
           name="author"
@@ -51,6 +48,7 @@
           class="form-control"
           id="pages"
           name="pages"
+          min="1"
           required
         />
         <label class="form-check-label">
--- a/debugging/book-library/index.html
+++ b/debugging/book-library/index.html
@@ -1,6 +1,6 @@
 <!DOCTYPE html>
 <html lang="">
-  <head>
+<head>
     <title>Book Library</title>
     <meta charset="utf-8"/>
     <meta name="viewport" content="width=device-width, initial-scale=1.0"/>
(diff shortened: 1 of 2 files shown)


debugging/book-library/index.html
```diff
@@ -51,6 +48,7 @@ <h1>Library</h1>
           class="form-control"
           id="pages"
           name="pages"
+          min="1"
           required
         />
```
cjyuan: Could you find out why the browser does not check if the value in this input field is a positive whole number (even if when the constraints `min="1"` is specified)?

debugging/book-library/index.html
cjyuan: Could you find out the tradeoff between 
A) specifying a callback via an `onclick` attribute in HTML, and
B) specifying a callback in JavaScript

?

debugging/book-library/script.js
```diff
@@ -1,103 +1,101 @@
 let myLibrary = [];
```

debugging/book-library/script.js
```diff
@@ -1,103 +1,101 @@
...
     return false;
   } else {
-    let book = new Book(title.value, title.value, pages.value, check.checked);
-    library.push(book);
+    let book = new Book(
+      titleInput.value,
+      authorInput.value,
+      Number(pagesInput.value),
+      checkInput.checked
+    );
```
cjyuan: - Is `Number(pagesInput.value)` always a valid positive whole number?

- The code checks the trimmed value but it uses the raw values.

    A better and safer approach to deal with user input is to first store the preprocessed input in variables, validate them, then use those cleaned values consistently throughout the rest of the code.

debugging/book-library/script.js
```diff
@@ -1,103 +1,101 @@
...
 }
 
-const title = document.getElementById("title");
-const author = document.getElementById("author");
-const pages = document.getElementById("pages");
-const check = document.getElementById("check");
+const titleInput = document.getElementById("title");
+const authorInput = document.getElementById("author");
+const pagesInput = document.getElementById("pages");
+const checkInput = document.getElementById("check");
```
cjyuan: Common practice is to declare all shared variables/constants at the beginning of the file before function definitions.

debugging/book-library/script.js
```diff
@@ -1,103 +1,101 @@
...
-    let deleteCell = row.insertCell(4);
-    titleCell.innerHTML = myLibrary[i].title;
-    authorCell.innerHTML = myLibrary[i].author;
-    pagesCell.innerHTML = myLibrary[i].pages;
+  myLibrary.forEach((book, i) => {
+    createBookRow(book, i);
+  });
+}
+
+const tableBody = document.querySelector("tbody");
```
cjyuan: Why not keep this variable declaration with the other variable declarations at the beginning of the file?

debugging/book-library/script.js
```diff
@@ -1,103 +1,101 @@
...
+let readStatus;
```
cjyuan: Does `readStatus` need to be shared among functions? If not, it is better to declare it as a local variable in the function where it is used.

debugging/book-library/script.js
```diff
@@ -1,103 +1,101 @@
...
+  changeBtn.className = "btn btn-success";
+
+  deleteBtn.className = "btn btn-warning";
+
+  if (!book.check) {
+    readStatus = "No";
+  } else {
+    readStatus = "Yes";
   }
+  changeBtn.innerText = readStatus;
```
cjyuan: This could be a good opportunity to practice using the `? :` conditional operator. Could you rewrite the code on lines 86–91 as a single statement?

debugging/book-library/script.js
```diff
@@ -1,103 +1,101 @@
...
+  changeBtn.addEventListener("click", function () {
+    book.check = !book.check;
+    render();
+  });
+  deleteBtn.addEventListener("click", function () {
+    alert(`You've deleted title: ${book.title}`);
+    myLibrary.splice(index, 1);
+    render();
```

In addition, `alert()` is a blocking function call. As a result, invoking it prevents the browser from updating the UI until the dialog is dismissed. 

Suggestion: Research for approaches that allows the UI to update before displaying the `alert` dialog.

debugging/book-library/index.html
d-odumosu: I think because i'm only checking if it is at least 1

debugging/book-library/index.html
d-odumosu: It's simple and straightforward,the behaviour is directly attached to the element, however we are mixing HTML with JavaScript and having (lots) of JavaScript inside HTML can become harder to maintain.

debugging/book-library/script.js
d-odumosu: changes from let to const

debugging/book-library/script.js
d-odumosu: I have moved them moved them, thank you

debugging/book-library/script.js
d-odumosu: I have also moved this

debugging/book-library/script.js
d-odumosu: I have implemented this, thank you

debugging/book-library/script.js
d-odumosu: i have moved it into the function

debugging/book-library/script.js
d-odumosu: i have implemented the ternery operator

debugging/book-library/script.js
d-odumosu: i have used the settimeout function for this, thank you

debugging/book-library/index.html
cjyuan, line 1: The formatting of the HTML code is not very consistent? Did you use any formatter to format the code?

debugging/book-library/script.js
```diff
@@ -1,103 +1,90 @@
...
+const tableBody = document.querySelector("tbody");
+const submitButton = document.querySelector('input[type="submit"]');
 
-window.addEventListener("load", function (e) {
+window.addEventListener("load", function () {
   populateStorage();
   render();
 });
 
+submitButton.addEventListener("click", handleSubmit);
```

debugging/book-library/script.js
```diff
@@ -1,103 +1,90 @@
-let myLibrary = [];
+const myLibrary = [];
+const titleInput = document.getElementById("title");
+const authorInput = document.getElementById("author");
+const pagesInput = document.getElementById("pages");
+const checkInput = document.getElementById("check");
+const tableBody = document.querySelector("tbody");
```
cjyuan: To be more specific, could you change the selector to "the tbody element inside the table with id='display'"?

cjyuan (commented):
Changes look good. I just have a question and some suggestions.

debugging/book-library/script.js
d-odumosu: i have implemented a setup function

debugging/book-library/index.html
d-odumosu: my HTML should format on save, i'll fix this

debugging/book-library/script.js
d-odumosu: implemented this suggestion, thank you

cjyuan (commented):
All good. Well done.

PR #535, opened 2026-08-11
London | 26-ITP-May | Khaliun Baatarkhuu | Sprint 2 | Book Library

khaliun-dev:
Self checklist

## Changelist

- Fixed books not rendering on page load
- Fixed error when submitting a new book
- Fixed author being saved as the title
- Fixed broken delete buttons
- Fixed incorrect read status
- Added validation for missing author
- Clear the form after successfully adding a book

## Testing

- Tested adding books
- Tested missing fields
- Tested read/unread status
- Tested deleting books
- Checked browser console for error

## Questions

None.

### DIFF
--- a/debugging/book-library/index.html
+++ b/debugging/book-library/index.html
@@ -31,7 +31,7 @@
       <div class="form-group">
         <label for="title">Title:</label>
         <input
-          type="title"
+          type="text"
           class="form-control"
           id="title"
           name="title"
@@ -39,7 +39,7 @@
         />
         <label for="author">Author: </label>
         <input
-          type="author"
+          type="text"
           class="form-control"
           id="author"
           name="author"
--- a/debugging/book-library/index.html
+++ b/debugging/book-library/index.html
@@ -65,7 +65,7 @@
           type="submit"
           value="Submit"
           class="btn btn-primary"
-          onclick="submit();"
+          onclick="addBook();"
         />
       </div>
     </div>
--- a/debugging/book-library/index.html
+++ b/debugging/book-library/index.html
@@ -1,20 +1,22 @@
 <!DOCTYPE html>
-<html>
+<html lang="en">
   <head>
-    <title> </title>
+    <title>My Book Library</title>
+    <meta charset="utf-8">
     <meta
-      charset="utf-8"
       name="viewport"
       content="width=device-width, initial-scale=1.0"
-    />
+    >
+
     <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.4.1/jquery.min.js"></script>
     <script src="https://cdnjs.cloudflare.com/ajax/libs/popper.js/1.16.0/umd/popper.min.js"></script>
     <script src="https://maxcdn.bootstrapcdn.com/bootstrap/4.4.1/js/bootstrap.min.js"></script>
+
     <link
       rel="stylesheet"
       href="https://maxcdn.bootstrapcdn.com/bootstrap/4.4.1/css/bootstrap.min.css"
-    />
-    <link rel="stylesheet" type="text/css" href="style.css" />
+    >
+    <link rel="stylesheet" type="text/css" href="style.css">
   </head>
 
   <body>
@@ -28,7 +30,7 @@
     </button>
 
     <div id="demo" class="collapse">
-      <div class="form-group">
+      <form id="bookForm" class="form-group">
         <label for="title">Title:</label>
         <input
           type="text"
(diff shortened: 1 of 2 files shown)


cjyuan (commented):

khaliun-dev:
@cjyuan i have made some improvements to the code from the general feedback, specifically:

## Changes

- Fixed book rendering on page load.
- Fixed book submission and author handling.
- Fixed Delete functionality.
- Fixed Read/No status.
- Added validation for missing fields.
- Clear the form after adding a book.
- Improved data types by storing pages as numbers.
- Improved DOM variable naming.
- Replaced `innerHTML` with `textContent`.
- Removed unnecessary button IDs.
- Simplified Read/No logic.
- Used `const` where variables aren't reassigned.

## Testing

Tested adding, validating, displaying, toggling Read/No, deleting books, and clearing the form. Browser console checked with no errors.

debugging/book-library/index.html
cjyuan, line 1: According to https://validator.w3.org/, there are errors in your `index.html`. Could you fix these errors?

- If possible, also make this input field to accept only positive whole numbers.

debugging/book-library/index.html
```diff
@@ -65,7 +65,7 @@ <h1>Library</h1>
           type="submit"
           value="Submit"
           class="btn btn-primary"
-          onclick="submit();"
+          onclick="addBook();"
```

debugging/book-library/script.js

debugging/book-library/script.js
```diff
@@ -7,42 +7,52 @@ window.addEventListener("load", function (e) {
...
-    render();
   }
-}
 
+  const book = new Book(
+    titleInput.value,
+    authorInput.value,
+    Number(pagesInput.value),
+    readInput.checked
+  );
```
cjyuan: - What if the input contains leading and trailing space characters?

- Could `Number(pagesInput.value)` evaluates to an invalid page count? What numbers should be considered "invalid page count"?

Note: A better and safer approach to deal with user input is to first preprocess/sanitise them, and then store the cleaned values in variables. Thereafter, refer only to the variables for cleaned values consistently throughout the rest of the code.

debugging/book-library/script.js
```diff
@@ -7,42 +7,52 @@ window.addEventListener("load", function (e) {
...
+
+  myLibrary.push(book);
+
+  titleInput.value = "";
+  authorInput.value = "";
+  pagesInput.value = "";
+  readInput.checked = false;
```
cjyuan: Note: If the input elements are inside a `<form>` element, we could also just call a form method to reset the form.

debugging/book-library/script.js
```diff
@@ -51,50 +61,54 @@ function Book(title, author, pages, check) {
...
-  let rowsNumber = table.rows.length;
-  //delete old table
-  for (let n = rowsNumber - 1; n > 0; n-- {
+  const table = document.getElementById("display");
+  const rowsNumber = table.rows.length;
+
+ 
+  for (let n = rowsNumber - 1; n > 0; n--) {
     table.deleteRow(n);
   }
```
cjyuan: Clearing a table by deleting its rows row by row is not efficient. Could you look up a more efficient way to clear a table (the `<tbody>` part of the table)?

debugging/book-library/script.js
```diff
@@ -51,50 +61,54 @@ function Book(title, author, pages, check) {
...
+
+    delButton.className = "btn btn-warning";
+    delButton.innerHTML = "Delete";
+
+    deleteCell.appendChild(delButton);
+
+    delButton.addEventListener("click", function () {
       alert(`You've deleted title: ${myLibrary[i].title}`);
       myLibrary.splice(i, 1);
       render();
```

---
`alert()` is a blocking function call. As a result, invoking it prevents the browser from updating the UI until the dialog is dismissed.

khaliun-dev:
@cjyuan I have addressed all the comments you have made. Thanks.

debugging/book-library/script.js
```diff
@@ -7,42 +7,61 @@ window.addEventListener("load", function (e) {
...
-    pages.value == ""
-  ) {
-    alert("Please fill all fields!");
-    return false;
-  } else {
-    let book = new Book(title.value, title.value, pages.value, check.checked);
-    library.push(book);
-    render();
-  }
+const bookForm = document.getElementById("bookForm");
```
cjyuan, line 28: Could consider placing all code that runs once on page load in a single function. For example, you could put it inside the page load callback or create a function named `init()` or `setup()` and call it once when the page loads.

debugging/book-library/script.js
```diff
@@ -7,42 +7,61 @@ window.addEventListener("load", function (e) {
...
+
     myLibrary.push(book1);
     myLibrary.push(book2);
-    render();
   }
 }
+const titleInput = document.getElementById("title");
+const authorInput = document.getElementById("author");
+const pagesInput = document.getElementById("pages");
+const readInput = document.getElementById("check");
```
cjyuan, line 26: Common practice is to declare all shared variables/constants at the beginning of the files, before function definitions. Doing so makes locating them easier.

debugging/book-library/script.js
```diff
@@ -7,42 +7,61 @@ window.addEventListener("load", function (e) {
...
+  const title = titleInput.value.trim();
+  const author = authorInput.value.trim();
+  const pages = Number(pagesInput.value);
+
+if (
+  title === "" ||
+  author === "" ||
+  !Number.isFinite(pages) ||
+  pages < 1 ||
+  !Number.isInteger(pages)
```
cjyuan, line 45: Could you figure out which check is redundant?

debugging/book-library/script.js
```diff
@@ -51,53 +70,55 @@ function Book(title, author, pages, check) {
...
+
+    deleteCell.appendChild(delButton);
+
+    delButton.addEventListener("click", function () {
+      const deletedTitle = myLibrary[i].title;
+
       myLibrary.splice(i, 1);
       render();
+
+      alert(`You've deleted title: ${deletedTitle}`);
```

cjyuan (commented):
Changes look good. Well done.

debugging/book-library/script.js
khaliun-dev: the code on line 43, due to another code preventing invalid values anyway.

PR #536, opened 2026-08-11
London | 26-ITP-May | Bisrat Tesfay | Sprint 2 | Book-Library

B3-Bisrat:
## Learners, PR Template

Self checklist

## Changelist

Fixed the library bugs, including adding, displaying, reading status, and deleting books. Also fixed form validation and author input issues.

### DIFF
--- a/debugging/book-library/index.html
+++ b/debugging/book-library/index.html
@@ -1,7 +1,7 @@
-<!DOCTYPE html>
+<!doctype html>
 <html>
   <head>
-    <title> </title>
+    <title>Book Library</title>
     <meta
       charset="utf-8"
       name="viewport"
@@ -31,7 +31,7 @@
       <div class="form-group">
         <label for="title">Title:</label>
         <input
-          type="title"
+          type="text"
           class="form-control"
           id="title"
           name="title"
@@ -39,7 +39,7 @@
         />
         <label for="author">Author: </label>
         <input
-          type="author"
+          type="text"
           class="form-control"
           id="author"
           name="author"
@@ -65,7 +65,7 @@
           type="submit"
           value="Submit"
           class="btn btn-primary"
-          onclick="submit();"
+          onclick="addBook()"
         />
       </div>
     </div>
--- a/debugging/book-library/index.html
+++ b/debugging/book-library/index.html
@@ -27,48 +27,44 @@
       Add new book
     </button>
 
-    <div id="demo" class="collapse">
-      <div class="form-group">
-        <label for="title">Title:</label>
-        <input
-          type="text"
-          class="form-control"
-          id="title"
-          name="title"
-          required
-        />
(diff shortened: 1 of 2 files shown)


debugging/book-library/index.html
cjyuan, line 1: According to https://validator.w3.org/, there are errors in your `index.html`. Could you fix these errors?

- For better user experience, you could also consider making this input field to accept only positive whole numbers.

debugging/book-library/script.js

debugging/book-library/index.html
```diff
@@ -65,7 +65,7 @@ <h1>Library</h1>
           type="submit"
           value="Submit"
           class="btn btn-primary"
-          onclick="submit();"
+          onclick="addBook()"
```

cjyuan (commented):
Can you also check if any of this general feedback can help you further improve your code?

B3-Bisrat:
Thank you for the feedback and comments. I’ve made the suggested changes based on your feedback. Could you please have a look and let me know if there’s anything else I should improve?

debugging/book-library/index.html
```diff
@@ -27,48 +27,44 @@ <h1>Library</h1>
...
-        <label for="pages">Pages:</label>
-        <input
-          type="number"
-          class="form-control"
-          id="pages"
-          name="pages"
-          required
-        />
-        <label class="form-check-label">
+    <form id="bookForm">
```
cjyuan, line 30: Could use lower kebab case for a naming consistency.

debugging/book-library/script.js
```diff
@@ -51,53 +58,53 @@ function Book(title, author, pages, check) {
...
+    deleteCell.appendChild(deleteButton);
+
+    deleteButton.addEventListener("click", function () {
+      const deletedTitle = myLibrary[i].title;
+
+      // delete first, then render
       myLibrary.splice(i, 1);
       render();
+
+      alert(`You've deleted title: ${deletedTitle}`);
```
cjyuan, line 107: `alert()` is a blocking function call. As a result, invoking it prevents the browser from updating the UI until the dialog is dismissed.

debugging/book-library/script.js
```diff
@@ -1,46 +1,53 @@
-let myLibrary = [];
+const myLibrary = [];
+
+const titleInput = document.getElementById("title");
+const authorInput = document.getElementById("author");
+const pagesInput = document.getElementById("pages");
+const checkInput = document.getElementById("check");
+const bookForm = document.getElementById("bookForm");
+
+bookForm.addEventListener("submit", addBook);
```
cjyuan, line 9: Could consider placing all code that runs once on page load in a single function. For example, you could put it inside the page load callback or create a function named `init()` or `setup()` and call it once when the page loads.

cjyuan (commented):
Changes look great. Well done.

PR #537, opened 2026-08-12
London | 26-ITP-May | Tomislav Dukez | Sprint 3 | Programmer Humour 

tomdu3:
## Learners, PR Template

Self checklist

## Changelist

Added Programmer Humour Task files
- `index.html`, `style.css`, and `scripts.js`
- I have also added a fallback image (default): `assets/no_image.webp`, while image is being fetched, or if API call fails


fetch/programmer-humour/script.js
```diff
@@ -0,0 +1,26 @@
+const imgEl = document.getElementsByTagName("img")[0];
```
hey-hammad: is there a better way of getting the specific image element without using the tag? 

Think about the following:

Right now we're assuming there is always 1 image element in the html, what if that changes to 0?

fetch/programmer-humour/script.js
```diff
@@ -0,0 +1,26 @@
...
+    }
+    const data = await response.json();
+    return data;
+  } catch (error) {
+    console.log("Error fetching data:", error);
+    return null;
+  }
+};
+
+let data;
```
hey-hammad: Is the outer `let data`; still needed here?

fetch/programmer-humour/script.js
```diff
@@ -0,0 +1,26 @@
...
+fetchData().then((data) => {
```
hey-hammad: Would it make sense to keep the async flow consistent by using async/await when consuming fetchData() as well, rather than mixing async/await and .then()?

Also, should we add a guard for the image element before setting imgEl.src?

hey-hammad (changes requested):
Nice work overall. I noticed a few code quality issues that would be good to clean up. Thanks

fetch/programmer-humour/script.js
tomdu3: I solved the problem with js element creation and fallbackImg integration which solves both problems

fetch/programmer-humour/script.js
tomdu3: removed data variable initialisation

fetch/programmer-humour/script.js
tomdu3: fixed fetch logic and guard for image

PR #538, opened 2026-08-12
London | 26-ITP-May | Dagim Daniel | Sprint 2 | book library

Dagim-Daniel:
## Learners, PR Template

Self checklist

## Changelist

fixing major bugs  of the book library project


debugging/book-library/index.html

- If possible, also make this input field to accept only positive whole numbers.

debugging/book-library/index.html
```diff
@@ -65,7 +65,7 @@ <h1>Library</h1>
           type="submit"
           value="Submit"
           class="btn btn-primary"
-          onclick="submit();"
+          onclick="submit()"
```

cjyuan (commented):

debugging/book-library/script.js

debugging/book-library/script.js
```diff
@@ -20,84 +24,88 @@ function populateStorage() {
...
 }
 
-function Book(title, author, pages, check) {
+function libraryBook(title, author, pages, isRead) {
   this.title = title;
   this.author = author;
   this.pages = pages;
-  this.check = check;
+  this.isRead = isRead;
 }
```
cjyuan: This is actually one of JS syntax to create a new type of objects. So by convention we should name it in Pascal Case.

debugging/book-library/script.js
```diff
@@ -20,84 +24,88 @@ function populateStorage() {
...
 
-const title = document.getElementById("title");
-const author = document.getElementById("author");
-const pages = document.getElementById("pages");
-const check = document.getElementById("check");
+const titleInput = document.getElementById("title");
+const authorInput = document.getElementById("author");
+const pagesInput = document.getElementById("pages");
+const readCheckbox = document.getElementById("check");
```
cjyuan: Common practice is to declare all shared variables at the beginning of the file, before function defintion.

debugging/book-library/script.js
```diff
@@ -20,84 +24,88 @@ function populateStorage() {
...
-    library.push(book);
-    render();
   }
+
+  let newBook = new libraryBook(
+    titleInput.value,
+    authorInput.value,
+    pagesInput.value,
+    readCheckbox.checked
+  );
```
cjyuan: - The code checks the trimmed value but it uses the untrimmed values.

- `pagesInput.value` is a string, and a number in scientific format may look unnatural as "number of pages".

debugging/book-library/script.js
```diff
@@ -20,84 +24,88 @@ function populateStorage() {
...
 function render() {
   let table = document.getElementById("display");
-  let rowsNumber = table.rows.length;
-  //delete old table
-  for (let n = rowsNumber - 1; n > 0; n-- {
+  let totalRows = table.rows.length;
+
+  for (let n = totalRows - 1; n > 0; n--) {
     table.deleteRow(n);
   }
```
cjyuan: Could you replace this with a more efficient code to clear a table (or clear the `<tbody>`) than deleting its rows one by one?

debugging/book-library/script.js
```diff
@@ -20,84 +24,88 @@ function populateStorage() {
...
-
-    changeBut.addEventListener("click", function () {
-      myLibrary[i].check = !myLibrary[i].check;
+
+    titleCell.textContent = myLibrary[i].title;
+    authorCell.textContent = myLibrary[i].author;
+    pagesCell.textContent = myLibrary[i].pages;
+
+    let toggleReadBtn = document.createElement("button");
+    toggleReadBtn.id = `readBtn-${i}`;
```
cjyuan: Why assign an ID attribute to the element?

debugging/book-library/script.js
```diff
@@ -20,84 +24,88 @@ function populateStorage() {
...
-    delBut.id = i + 5;
-    deleteCell.appendChild(delBut);
-    delBut.className = "btn btn-warning";
-    delBut.innerHTML = "Delete";
-    delBut.addEventListener("clicks", function () {
+    deleteBtn.addEventListener("click", function () {
       alert(`You've deleted title: ${myLibrary[i].title}`);
       myLibrary.splice(i, 1);
       render();
     });
```

---
Please note that `alert()` is a blocking function call. As a result, invoking it prevents the browser from updating the UI until the dialog is dismissed.

cjyuan (commented):
Changes look good. Well done.

PR #539, opened 2026-08-12
Birmingham | 26-ITP-May | Gabriel Pawuoi | Sprint 2 | Book Library

KhotKeys:
## Learners, PR Template

Self checklist

## Changelist

Fixed all bugs in the book library, specifically:

- fixed a syntax error in the for loop in the render function.
- fixed the submit function using the wrong variable name (library instead of myLibrary).
- fixed the author field using title value instead of author value.
- fixed the delete button variable name mismatch (delBut vs delButton) and wrong event name (clicks instead of click).
- fixed the inverted read status logic showing wrong Yes/No values.

## Questions
None.


cjyuan (commented):
Could you check if any of this general feedback can help you further improve your code?

debugging/book-library/index.html
```diff
@@ -28,46 +27,54 @@ <h1>Library</h1>
...
-        <input
-          type="number"
-          class="form-control"
-          id="pages"
-          name="pages"
-          required
-        />
-        <label class="form-check-label">
+      <!-- Refactor: Form wrapped to allow formal submit handling and native constraint execution -->
+      <form id="book-form" onsubmit="event.preventDefault(); window.submit();">
```
cjyuan: Could you look up the trade-off between
- attaching event listener in JS via `.addEventListener()`

debugging/book-library/script.js
```diff
@@ -1,46 +1,47 @@
...
 
-const title = document.getElementById("title");
-const author = document.getElementById("author");
-const pages = document.getElementById("pages");
-const check = document.getElementById("check");
+// Refactor: Updated variable names with descriptive suffixes to clearly reflect that they store DOM nodes
+const titleInput = document.getElementById("title");
+const authorInput = document.getElementById("author");
+const pagesInput = document.getElementById("pages");
+const checkInput = document.getElementById("check");
```
cjyuan: Note: Common practice is to declare all shared variables at the beginning of the file, before function definition.

debugging/book-library/script.js
```diff
@@ -1,46 +1,47 @@
...
+
+  let book = new Book(formattedTitle, formattedAuthor, pageCount, checkInput.checked);
+  myLibrary.push(book);
+  render();
+
+  // Clear inputs on success
+  titleInput.value = "";
+  authorInput.value = "";
+  pagesInput.value = "";
+  checkInput.checked = false;
```
cjyuan: Good idea to clear the form. We could also call a form's method to reset the form.

debugging/book-library/script.js
```diff
@@ -51,53 +52,53 @@ function Book(title, author, pages, check) {
...
+    deleteCell.appendChild(deleteBtn);
+    deleteBtn.className = "btn btn-warning btn-sm";
+    deleteBtn.textContent = "Delete";
+    
+    deleteBtn.addEventListener("click", function () {
+      const targetTitle = myLibrary[i].title;
+      // Refactor: Resolved timing bug. Data mutation and UI render occur safely BEFORE the blocking window.alert runs
       myLibrary.splice(i, 1);
       render();
+      alert(`You've successfully deleted title: ${targetTitle}`);
```

cjyuan (commented):
Changes look good.

I only have a few suggestions.

debugging/book-library/script.js
KhotKeys: Thank you very much, this is worth considering for my future practices, I am thankful

KhotKeys: It is fixed

debugging/book-library/script.js
KhotKeys: Worth it, it is fixed

debugging/book-library/script.js
KhotKeys: I will surely adhere to this, thanks

KhotKeys: I have refactored the deletion handler to wrap the alert() in a setTimeout(..., 0), which yields control back to the browser event loop and allows the UI to update immediately before the dialog blocks execution.

debugging/book-library/index.html
KhotKeys: I have revisited and fixed it

debugging/book-library/script.js
```diff
@@ -1,47 +1,44 @@
...
     render();
   }
 }
 
-const title = document.getElementById("title");
-const author = document.getElementById("author");
-const pages = document.getElementById("pages");
-const check = document.getElementById("check");
+// 2. Attach event listener in JS via .addEventListener() instead of HTML inline
+bookForm.addEventListener("submit", function (event) {
```
cjyuan: Could consider 
- Define the "on submit" callback function as a named function.
- Place the "callback setup" statement together with all the code that run once on page load in a single function. For example, you could put it inside the page load callback or create a function named `init()` or `setup()` and call it once when the page loads.

Doing so could make it easier to locate and manage all the code that runs once when the app starts.

cjyuan (commented):
Changes look good.

debugging/book-library/script.js
KhotKeys: Perfect, let me fall back and add those, thank you so much

> Could consider
> 
> * Define the "on submit" callback function as a named function.
> * Place the "callback setup" statement together with all the code that run once on page load in a single function. For example, you could put it inside the page load callback or create a function named `init()` or `setup()` and call it once when the page loads.
> 
> Doing so could make it easier to locate and manage all the code that runs once when the app starts.

Perfect, let me fall back and add those, thank you so much

PR #544, opened 2026-08-12
Birmingham | 26-ITP-May | Gabriel Pawuoi | Sprint 3 | Programmer Humour

KhotKeys:
## Learners, PR Template

Self checklist

## Changelist

Learnt about fetch and the DOM, specifically:
- used fetch to make an API call to the xkcd endpoint and rendered the comic image into the DOM.


fetch/programmer-humour/script.js
```diff
@@ -0,0 +1,18 @@
+function getLatestComic() {
+  fetch("https://xkcd.now.sh/?comic=latest")
+    .then(function (response) {
+      if (!response.ok) throw new Error("Network response was not ok");
+      return response.json();
+    })
+    .then(function (data) {
+      console.log(data);
+      document.getElementById("comic-img").src = data.img;
```
hey-hammad: would it be better to if we cache the comic-img element instead of querying the DOM twice? It would make the code a bit cleaner and avoid repeating the selector.

hey-hammad: Also can we assume that elements comic-img and error are always available, and the absence will not break the code?

hey-hammad (changes requested):
Good implementation overall. I have a few suggestions around DOM element checks, avoiding repeated DOM queries. Thanks

fetch/programmer-humour/script.js
KhotKeys: > would it be better to if we cache the comic-img element instead of querying the DOM twice? It would make the code a bit cleaner and avoid repeating the selector.

All changes are fully fixed, thanks for the feedback

PR #546, opened 2026-08-12
London | 26-ITP-May | Martin Mwaka | Sprint 3 | Programmer humour

Temceo:
Self checklist

## Changelist

Completed tasks for programmer humour

### DIFF
--- a/fetch/programmer-humour/script.js
+++ b/fetch/programmer-humour/script.js
@@ -37,13 +37,11 @@ async function setup() {
       loadingMessage.remove();
     }
   } catch (error) {
-    console.error("Failed to load image:", error);
     showMessage("Sorry, we could not load the image right now.");
   }
 }
 
 function renderImage() {
-  console.log(state.imgData);
   if (imageElement && state.imgData.img) {
     imageElement.src = state.imgData.img;
     imageElement.alt = state.imgData.title || "Comic image";
--- a/fetch/programmer-humour/script.js
+++ b/fetch/programmer-humour/script.js
@@ -1,7 +1,6 @@
 const imageElement = document.querySelector("img");
 const IMG_URL = `https://xkcd.now.sh/?comic=latest`;
 
-let cachedImage = null;
 let imageCache = null;
 let imagePromise = null;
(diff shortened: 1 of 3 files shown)


fetch/programmer-humour/script.js
```diff
@@ -0,0 +1,64 @@
+const imageElement = document.querySelector("img");
+const endpoint = `https://xkcd.now.sh/?comic=latest`;
+
+let imageCache = null;
```
hey-hammad: Do we expect fetchImageData() to be called multiple times during the page lifecycle? If not, could we simplify the implementation by removing the cache/promise management and fetching the data directly?

fetch/programmer-humour/script.js
```diff
@@ -0,0 +1,64 @@
...
+let imagePromise = null;
+
+const state = {
```
hey-hammad: Is there a reason we're storing the response in a global state object? Since the data appears to be consumed immediately by renderImage(), could we pass it directly as a parameter instead of maintaining additional state?

fetch/programmer-humour/script.js
```diff
@@ -0,0 +1,64 @@
...
+  imgData: {},
+};
+
+async function fetchImageData() {
```
hey-hammad: I noticed that function is defined using async, but its using .then() inside. Would it make sense to use await inside fetchImageData() as well?

hey-hammad (changes requested):
Good structure overall, but the solution seems more complex than necessary for a single fetch-on-load scenario.

fetch/programmer-humour/script.js
Temceo: I have simplified this to collect the data directly

fetch/programmer-humour/script.js
Temceo: I am now passing the data directly to renderImage

fetch/programmer-humour/script.js
Temceo: I am now only using await

PR #549, opened 2026-08-13
London | 26-ITP-May | Rizqah Popoola | Sprint 2 |Book Library

risikatpopoola:
## Learners, PR Template

Self checklist

## Changelist

In this PR, i have fixed all the bugs in script.js and refactored the script to be more efficient. I have also validated the html code to ensure it's free of errors.

### DIFF
--- a/debugging/book-library/index.html
+++ b/debugging/book-library/index.html
@@ -51,21 +51,21 @@
           class="form-control"
           id="pages"
           name="pages"
+          min="1"
+          step="1"
           required
         >
-        <label class="form-check-label">
-          <input
-            type="checkbox"
-            class="form-check-input"
-            id="check"
-            value=""
-          >Read
+        <div class="form-check">
+        <input class="form-check-input" type="checkbox" id="check">
+        <label class="form-check-label" for="check">
+        Read
         </label>
+        </div>
         <input
           type="submit"
           value="Submit"
           class="btn btn-primary"
-          onclick="submit();"
+          id ="submit"
         >
       </div>
     </div>
--- a/debugging/book-library/index.html
+++ b/debugging/book-library/index.html
@@ -1,20 +1,20 @@
 <!DOCTYPE html>
-<html>
+<html lang ='en'>
   <head>
-    <title> </title>
+    <title>Rizqah's Book Library</title>
+<meta charset="utf-8">
     <meta
-      charset="utf-8"
       name="viewport"
       content="width=device-width, initial-scale=1.0"
-    />
+    >
     <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.4.1/jquery.min.js"></script>
     <script src="https://cdnjs.cloudflare.com/ajax/libs/popper.js/1.16.0/umd/popper.min.js"></script>
     <script src="https://maxcdn.bootstrapcdn.com/bootstrap/4.4.1/js/bootstrap.min.js"></script>
     <link
       rel="stylesheet"
       href="https://maxcdn.bootstrapcdn.com/bootstrap/4.4.1/css/bootstrap.min.css"
-    />
-    <link rel="stylesheet" type="text/css" href="style.css" />
+    >
+    <link rel="stylesheet" type="text/css" href="style.css">
   </head>
 
   <body>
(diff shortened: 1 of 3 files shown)


debugging/book-library/index.html
```diff
@@ -31,45 +31,45 @@ <h1>Library</h1>
...
           required
-        />
+        >
         <label for="pages">Pages:</label>
         <input
           type="number"
           class="form-control"
           id="pages"
           name="pages"
           required
```
cjyuan: Why not make this input field to accept only positive whole number?

debugging/book-library/index.html
```diff
@@ -31,45 +31,45 @@ <h1>Library</h1>
...
-        />
+        >
         <label class="form-check-label">
           <input
             type="checkbox"
             class="form-check-input"
             id="check"
             value=""
```

debugging/book-library/script.js
```diff
@@ -1,46 +1,58 @@
 let myLibrary = [];
```
cjyuan: Could we declare `myLibrary` in a way that prevents it from being accidentally reassigned?

debugging/book-library/index.html
```diff
@@ -31,45 +31,45 @@ <h1>Library</h1>
...
-          />Read
+          >Read
         </label>
         <input
           type="submit"
           value="Submit"
           class="btn btn-primary"
           onclick="submit();"
```
- attaching event listener usng `.addEventListener()` in JS?

debugging/book-library/script.js
```diff
@@ -1,46 +1,58 @@
...
+  const pages = Number(pagesInput.value);
+
+  if (pages <= 0 || !Number.isInteger(pages)) {
+    alert("Please enter a valid page number.");
+    return false;
+  }
+
+  const book = new Book(
+    titleInput.value.trim(),
+    authorInput.value.trim(),
```
cjyuan: A better and safer approach to deal with user input is to first preprocess/sanitise them, and then store the cleaned values in variables. Thereafter, refer only to the variables for cleaned values consistently throughout the rest of the code.

cjyuan (commented):
Solid start.

I think there are some improvements you could still make.

debugging/book-library/script.js
risikatpopoola: I have now declared this using const

debugging/book-library/index.html
risikatpopoola: It is easier to manage event listeners in js and it provides more control.

debugging/book-library/index.html
```diff
@@ -31,45 +31,45 @@ <h1>Library</h1>
...
-        />
+        >
         <label for="pages">Pages:</label>
         <input
           type="number"
           class="form-control"
           id="pages"
           name="pages"
+          min="1"
+          step="1"
```
cjyuan: The browser will only check the input against these constraints when the user submit a form. Currently these input elements are not inside a form.

debugging/book-library/index.html
```diff
@@ -31,45 +31,45 @@ <h1>Library</h1>
...
-        <label class="form-check-label">
-          <input
-            type="checkbox"
-            class="form-check-input"
-            id="check"
-            value=""
-          />Read
+        >
+        <div class="form-check">
+        <input class="form-check-input" type="checkbox" id="check">
```
cjyuan: Why not use a code formatter to keep the code consistently formatted?

debugging/book-library/script.js
```diff
@@ -51,53 +57,63 @@ function Book(title, author, pages, check) {
...
+    const pagesCell = row.insertCell(2);
+    const wasReadCell = row.insertCell(3);
+    const deleteCell = row.insertCell(4);
+
+    titleCell.textContent = myLibrary[i].title;
+    authorCell.textContent = myLibrary[i].author;
+    pagesCell.textContent = myLibrary[i].pages;
+    const submitButton = document.getElementById("submit");
+
+    submitButton.addEventListener("click", submit);
```
cjyuan: Why attach the event listener to the submit button here?

risikatpopoola: I completely removed the event listener for the submit button and added it to the book form submission instead.

cjyuan: **How many times** do you expect the app to add the event listeners to the submit button?

cjyuan:
Please note that if you have made any changes, you need to push them to GitHub.

debugging/book-library/script.js
risikatpopoola: It should only be once, i have now taken it out of setup

debugging/book-library/script.js
```diff
@@ -115,5 +115,5 @@ function render() {
     });
   }
 }
-
+bookForm.addEventListener("submit", addBook);
```
cjyuan, line 118: Could consider placing all code that runs once on page load in a single function. For example, you could put it inside the page load callback or create a function named `init()` or `setup()` and call it once when the page loads.

cjyuan (commented):
Changes look good.

PR #552, opened 2026-08-14
London | 26-ITP-May | Edina Kurdi | Sprint 2 | Book Library

edinakurdi:
## Learners, PR Template

Self checklist

## Changelist

- Added the missing closing bracket in the render() function's loop 
- Corrected the array name to myLibrary and fixed the event listener to use click 
- Fixed the variable mapping in render() 
- Fixed variable name mismatch. (delButton) 
- Fixed the submit() function so it saves the actual author's name instead of copying the title twice. (Note - in the commit msg I may have accidentally said this was in render()) 
- Fixed the read (y/n) display logic for new table entries 

Other corrections I made: 
Added missing author form validation, fixed HTML input types, corrected a book title typo, and removed redundant CSS.


debugging/book-library/index.html
```diff
@@ -1,96 +1,63 @@
...
-        />
-      </div>
+  <div id="demo" class="collapse">
+    <div class="form-group">
+      <label for="title">Title:</label>
+      <input type="text" class="form-control" id="title" name="title" required />
+      <label for="author">Author: </label>
+      <input type="text" class="form-control" id="author" name="author" required />
+      <label for="pages">Pages:</label>
+      <input type="number" class="form-control" id="pages" name="pages" required />
```
cjyuan: Why not make this input element to accept only positive whole number?

debugging/book-library/index.html
```diff
@@ -1,96 +1,63 @@
...
+      <label class="form-check-label">
+        <input type="checkbox" class="form-check-input" id="check" value="" />Read
+      </label>
```
cjyuan, line 35: If this checkbox is not showing, could you try fixing the issue with the help of AI.  The issue is related to Bootstrap 4.4.1. So mentioning "Bootstrap 4.4.1" to AI might help.

debugging/book-library/index.html
```diff
@@ -1,96 +1,63 @@
...
+      <input type="submit" value="Submit" class="btn btn-primary" onclick="submit();" />
```
cjyuan: Could you look up the trade-off between 
- attaching event listener in JS using `.addEventListener()`?

cjyuan (commented):

debugging/book-library/index.html
edinakurdi: inline breaks separation of concersn and doesnt allow multiple event listeneres, and forces the function to global scope which can cause issues, but it is quick to write, shorter
addEventListener supports multiple event listeners, no global scope, but requires more code. 
updated

debugging/book-library/script.js
```diff
@@ -7,41 +7,42 @@ window.addEventListener("load", function (e) {
...
-    pages.value == null ||
-    pages.value == ""
-  ) {
-    alert("Please fill all fields!");
-    return false;
-  } else {
-    let book = new Book(title.value, title.value, pages.value, check.checked);
-    library.push(book);
-    render();
+bookForm.addEventListener("submit", function (e) {
```
cjyuan, line 29: Could consider placing all code that runs once on page load in a single function. For example, you could put it inside the page load callback or create a function named `init()` or `setup()` and call it once when the page loads.

debugging/book-library/script.js
```diff
@@ -51,53 +52,47 @@ function Book(title, author, pages, check) {
...
+
+    deleteBtn.addEventListener("click", function () {
+      const confirmDel = confirm(
+        `Are you sure you want to delete "${myLibrary[i].title}"?`
+      );
+      if (confirmDel) {
+        myLibrary.splice(i, 1);
+        render();
+      }
     });
```
cjyuan, line 96: Informing the user that the book has been deleted is not quite the same as asking whether they want to delete the book though.

cjyuan (commented):
Changes look good.

Well done.

PR #553, opened 2026-08-14
Glasgow | 26-ITP-May | Francesco Romano Monda | Sprint 2 | Book Library

fromonda:
## Self checklist

## Changelist

- Fixed the initial book list so existing books are displayed on page load.
- Fixed the error occurring when adding a new book.
- Corrected author handling so the book’s author is saved/displayed correctly.
- Fixed the delete functionality so books can be removed from the list.
- Fixed the read status so it correctly reflects the value selected when adding a book.
- Added validation to prevent books with missing title, author, page count, or read status from being added.
- Added an alert when required book information is missing.
- Fixed other minor issues affecting book list functionality and data handling.


debugging/book-library/index.html
cjyuan: If this checkbox is not showing, could you use an AI tool to find out how to fix the issue?
The issue is related to Bootstrap 4.4.1, so mentioning "Bootstrap 4.4.1" could help.

cjyuan, line 58: Why not make this input element to accept only positive whole number?

cjyuan (commented):

debugging/book-library/index.html
fromonda: I have updated the pages input to accept only positive whole numbers by setting min="1" and step="1" on the number input.

fromonda: I updated the checkbox using the appropriate Bootstrap form-check, form-check-input, and form-check-label classes. I also checked that the checkbox is displaying correctly.

fromonda:
> Could you also check if any of this general feedback can help you further improve your code? https://github.com/CodeYourFuture/Module-Data-Flows/blob/general-review-feedback/debugging/book-library/feedback.md
> 
> Doing so can help me speed up the review process. Thanks.

Thank you for pointing me towards the improvement guidelines and for trying to speed up the review process. I put all the improvements indicated into practice, although it took some time. I hope the refactoring, input validation, and other changes are better now.

debugging/book-library/index.html
```diff
@@ -52,24 +51,19 @@ <h1>Library</h1>
           id="pages"
           name="pages"
           required
+          min="1"
+          step="1"
         />
```
cjyuan, line 58: Even with these constraints, currently the browser would still allow a user to submit value like `12.345` or `-100`. Could you find out why and then fix the issue?

debugging/book-library/script.js
```diff
@@ -1,48 +1,49 @@
...
+const book2 = new Book(
+  "The Old Man and the Sea",
+  "Ernest Hemingway",
+  127,
+  true
+);
+
+myLibrary.push(book1);
+myLibrary.push(book2);
+render();
```

debugging/book-library/script.js
```diff
@@ -1,48 +1,49 @@
...
-    pages.value == ""
-  ) {
-    alert("Please fill all fields!");
-    return false;
-  } else {
-    let book = new Book(title.value, title.value, pages.value, check.checked);
-    library.push(book);
-    render();
+  if (!title || !author || !Number.isInteger(pages) || pages < 1) {
+    showMessage("Please enter valid book information.", "danger");
```
cjyuan, line 49: When I clicked the Submit button without any input, I couldn't see any error message.

cjyuan (commented):
Code looks pretty solid. 

There seems to be a bug.

debugging/book-library/index.html
fromonda: I tested this again and found that decimal and negative page values are now rejected and the error message is displayed. It is possible that this issue occurred before I added the error message.

debugging/book-library/script.js
fromonda: I have updated my code by creating an init() function and moving the code that runs when the application starts into it.

debugging/book-library/script.js
fromonda: I found the issue: I had not included an element with id="message" in my HTML. I added the missing element, and the validation error message now displays correctly when the Submit button is clicked without any input.

fromonda:
> Code looks pretty solid.
> 
> There seems to be a bug.

This issue may have been related to the changes I made while fixing the validation and error-message handling.

debugging/book-library/script.js
```diff
@@ -1,29 +1,42 @@
 const myLibrary = [];
 
-// Get elements from HTML
-const titleInput = document.getElementById("title");
-const authorInput = document.getElementById("author");
-const pagesInput = document.getElementById("pages");
-const checkInput = document.getElementById("check");
-const submitBtn = document.getElementById("submitBtn");
-const bookList = document.getElementById("bookList");
-const messageEl = document.getElementById("message");
```
cjyuan, line 10: This approach is actually safer and preferred. Declaring these shared variables using `let` is too risky.

debugging/book-library/script.js
cjyuan: This was the bug I wanted you to find. You fixed it!

debugging/book-library/index.html
cjyuan: The error message you mentioned is generated by your JS code, and not by the browser.
The browser will only check the input against those constraints when a user submits a form. Currently those input elements are not inside a form, and that's why the browser does not enforce the constraints.

cjyuan:
I will mark this PR as "Complete" first.

PR #558, opened 2026-08-15
London | 26-ITP-May | Gideon Defar | Sprint 2 | Book Library

gideondefar:
Self checklist

## Changelist

In this PR, i have fixed all the bugs in script.js and refactored the script to be more efficient. I have also validated the html code to ensure it's free of errors.

### DIFF
--- a/debugging/book-library/script.js
+++ b/debugging/book-library/script.js
@@ -101,11 +101,12 @@ function render() {
     deleteButton.addEventListener("click", function () {
       const deletedTitle = myLibrary[i].title;
 
-      // delete first, then render
       myLibrary.splice(i, 1);
       render();
 
-      alert(`You've deleted title: ${deletedTitle}`);
+      setTimeout(() => {
+        alert(`You've deleted title: ${deletedTitle}`);
+      }, 0);
     });
   }
 }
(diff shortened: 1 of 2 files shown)


debugging/book-library/script.js
```diff
@@ -1,46 +1,53 @@
...
+const pagesInput = document.getElementById("pages");
+const checkInput = document.getElementById("check");
+const bookForm = document.getElementById("bookForm");
+
+bookForm.addEventListener("submit", addBook);
 
 window.addEventListener("load", function (e) {
   populateStorage();
   render();
 });
```

debugging/book-library/script.js
```diff
@@ -51,53 +58,53 @@ function Book(title, author, pages, check) {
...
+    deleteCell.appendChild(deleteButton);
+
+    deleteButton.addEventListener("click", function () {
+      const deletedTitle = myLibrary[i].title;
+
+      // delete first, then render
       myLibrary.splice(i, 1);
       render();
+
+      alert(`You've deleted title: ${deletedTitle}`);
```

cjyuan (commented):
According to https://validator.w3.org/, there are errors in your `index.html`. Could you fix these errors?

Other changes look good. You did your homework!

debugging/book-library/script.js
gideondefar: Allow the browser to update the UI before showing the blocking alert

cjyuan (commented):
Changes look good. Well done.

PR #562, opened 2026-08-17
London | 26-ITP-May | Russom Gebremeskel | Sprint 2 | Book Library 

russom-g:
## Learners, PR Template

Self checklist

## Changelist

Bug fixed. Dead code removed.


cjyuan (commented):
Could consider placing all code that runs once on page load in a single function. For example, you could put it inside the page load callback or create a function named `init()` or `setup()` and call it once when the page loads.

russom-g:
Thanks for the feedback. Now a setup function is implemented. Both populateStorage and render function are wrapped inside the function.

cjyuan (commented):
- Could you remove `.vscode/settings.json` and revert the changes made in `.gitignore`? Otherwise the validation bot will reject this PR.

- Could you also check if any of this general feedback can help you further improve your code?

russom-g:
Files that shouldn't be here are removed. I also have made some improvements following the requirement document. Thanks

debugging/book-library/index.html
```diff
@@ -31,45 +28,47 @@ <h1>Library</h1>
...
         <label for="pages">Pages:</label>
         <input
           type="number"
           class="form-control"
           id="pages"
           name="pages"
+          min="1"
           required
-        />
+        >
```
cjyuan, line 53: The browser checks the input elements against the specified constraints only when a user submits a form.
Without `<form>`, the browser won't check the input against the constraints such as `required` and `min="1"`.

debugging/book-library/index.html
```diff
@@ -31,45 +28,47 @@ <h1>Library</h1>
...
         <label class="form-check-label">
           <input
             type="checkbox"
             class="form-check-input"
             id="check"
-            value=""
-          />Read
+          >
```
cjyuan, line 59: This checkbox is not showing.

debugging/book-library/script.js
```diff
@@ -1,103 +1,92 @@
...
+    return;
   }
-}
-
-function Book(title, author, pages, check) {
-  this.title = title;
-  this.author = author;
-  this.pages = pages;
-  this.check = check;
+  if (isNaN(pages) || pages <= 0) {
```
cjyuan, line 36: Could consider check if `pages` is a whole number instead.

cjyuan:
note: You shouldn't have deleted the `.gitignore` that was in the original repo.

